PR #2, opened 2015-08-01
Fix grammar bug

andresgottlieb:


sindresorhus:
I guess I need a linter for English too :p Thanks!

PR #25, opened 2015-09-11
pass options to lintText from cli

### LINKED ISSUE #24: Other options are ignored when using `--stdin`
For example:
```
# echo "let foo; console.log(foo);" | xo --stdin --esnext

<text>
  1:2  error  Unexpected token let
```
Looks like the cli isn't passing the options to `lintText()`:
https://github.com/sindresorhus/xo/blob/master/cli.js#L74

floatdrop:
Closes #24


sindresorhus:
Thanks :)

Issue #28, opened 2015-09-21
Vim support


sindresorhus:
Someone needs to do a Syntastic plugin.

https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide

arthurvr:
There we go: https://github.com/arthurvr/syntastic/commit/289983577a9d7e0a1959da39d13c6f04f5d63022. Just gonna run it locally for a few days before doing the PR as I feel like I've missed some more edge cases.

arthurvr:
(Some more people to test it definitely welcome!)

PR #29, opened 2015-09-22
Add documentation about using XO with Vim

arthurvr:
Ref https://github.com/scrooloose/syntastic/pull/1561
Fixes gh-28

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -197,6 +197,17 @@ XO is based on ESLint. This project started out as just a shareable ESLint confi
 - [Sublime Text](https://github.com/sindresorhus/SublimeLinter-contrib-xo)
 - [Atom](https://github.com/sindresorhus/atom-linter-xo)
 
+### Vim
+
+You can use [Syntastic](https://github.com/scrooloose/syntastic)'s ESLint checker with the following settings in your `.vimrc` file:
+
+```vim
+let g:syntastic_javascript_eslint_generic = 1
+let g:syntastic_javascript_eslint_exec = 'xo'
+let g:syntastic_javascript_eslint_args = '--compact'
+let g:syntastic_javascript_checkers = ['eslint']
+```
+
 
 ## Build-system plugins
 


readme.md
```diff
@@ -197,6 +197,17 @@ XO is based on ESLint. This project started out as just a shareable ESLint confi
...
+### Vim
+
+You can use [Syntastic](https://github.com/scrooloose/syntastic)'s ESLint checker with the following settings in your `.vimrc` file:
+
+```vim
+let g:syntastic_javascript_eslint_generic = 1
+let g:syntastic_javascript_eslint_exec = 'xo'
+let g:syntastic_javascript_eslint_args = '--compact'
+let g:syntastic_javascript_checkers = ['eslint']
+```
```
rauchg: I'm curious @sindresorhus, why is this not like `standard`, just a `checker`?

https://github.com/feross/standard#vim

Qix-: @rauchg Because XO is just a set of rules for eslint.

rauchg: @Qix- it was confusing to me because it somehow implies that it _always_ execs `xo`.

Qix-: @rauchg Pretty sure it does, though having both `eslint` and `xo` wouldn't make sense either since they're both going to try to lint javascript files and give you weird results anyway. ViM isn't going to be smart enough to figure out which scripts you want linted as XO and which you want linted as generic ESLint.

rauchg: I have a $PATH that points to `./node_modules/.bin`, so I could potentially list 3 programs in order of precedence to check for:

`['xo', 'standard', 'eslint']`

rauchg: This seems to me the most elegant solution as it doesn't make any global assumptions

Qix-: And `xo` will always win because its checks to see if the current buffer is viable will always succeed since they're the same as eslint and probably pretty similar to standard. What will Syntastic do to determine which linter to use?

rauchg: Not finding it in `$PATH`

rauchg: Not sure that's what Synctastic does, purely theoretical :P

Qix-: Even by that logic, if `XO` is in your path, it will always win regardless of what else is in the the list of checkers.

rauchg: Right, but it shouldn't be in the path if the project is not configured with `xo`. Again, this would be optimized for the (in my opinion _correct_) usage of exclusively local per-project linters.

Qix-: @rauchg What is adding `xo` to your path based on your current directory?

rauchg: A relative path entry in `$PATH` (`./node_modules/.bin`)

Qix-: Not sure that's something we want to support tbh. You're doing two things that are edge-casey and not really intended functionality: using `PATH` with relative paths, and then using syntastic's chaining ability to select the linter you want to use. While it works fine in _your_ environment that's not something I think will work globally in _everyone's_ environment.

For clarity, from the Syntastic docs:

> [Multiple checkers tells] syntastic to run the php checker first, and if no errors are found, run phpcs, and then phpmd[, etc.].

rauchg: > You're doing two things that are edge-casey and not really intended functionality

Are relative paths not intended functionality, or just maybe less common? For the record, they appear to be relatively (no pun intended) uncommon:

https://twitter.com/rauchg/status/787811383643475969

rauchg: So, it seems like what I'm going to do is have a `javascript-lint` wrapper that checks for the right linter to use, depending on what it finds in `./node_modules/.bin`.

That said, it's still desirable I think to have a `eslint`-compatible binary like `standard` does ?

rauchg: I agree completely with @sindresorhus comments there.

sindresorhus: Actually, it's now possible to create external linters. See: https://github.com/sindresorhus/xo/issues/160 Help welcome.

PR #34, opened 2015-10-04
update packages, use new eslint and xo-init

mattbanks:


shinnn:
Not needed. https://nodesource.com/blog/semver-tilde-and-caret

PR #35, opened 2015-10-06
Add support for plugins

### LINKED ISSUE #30: How to integrate ESLint plugins?
I was just wondering the best way to integrate ESLint plugins to XO.

Let's say I want to use [`eslint-plugin-no-use-extend-native`](https://github.com/dustinspecker/eslint-plugin-no-use-extend-native/) along with XO, what should I do? Does XO simply invokes ESLint and things work as expected or is there any trick that I need to do?

kevva:
Don't know if you had another API in mind, but this at least does the job. If you want, I can include https://github.com/dustinspecker/eslint-plugin-no-use-extend-native/ by default too (like mentioned in #30).

Fixes #30.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -20,6 +20,7 @@ var cli = meow({
 		'  --ignore        Additional paths to ignore  [Can be set multiple times]',
 		'  --space         Use space indent instead of tabs  [Default: 2]',
 		'  --no-semicolon  Prevent use of semicolons',
+		'  --plugin        Include third-party plugins  [Can be set multiple times]',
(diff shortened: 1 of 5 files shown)


rafaelrinaldi:
LGTM! Thanks for the PR.

silvenon:
I really miss this :+1:

sindresorhus:
> If you want, I can include https://github.com/dustinspecker/eslint-plugin-no-use-extend-native/ by default too (like mentioned in #30).

:+1:

PR #41, opened 2015-10-20
Use `no-use-extend-native` plugin by default

kevva:
Prevents use of extended native objects.

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,6 +66,8 @@ function handleOpts(opts) {
 		opts._config.baseConfig = 'xo/esnext';
 	}
 
+	opts._config.plugins.push('no-use-extend-native');
+
 	return opts;
 }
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -66,6 +66,8 @@ function handleOpts(opts) {
 		opts._config.baseConfig = 'xo/esnext';
 	}
 
+	opts._config.plugins.push('no-use-extend-native');
```
kevva: Did the easy thing and placed it here, but would you like me to set it in `DEFAULT_CONFIG` instead?

sindresorhus: No, this is fine.

sindresorhus:
Can you set the rule config here too? I don't want to put it in https://github.com/sindresorhus/eslint-config-xo as the user would have to install it manually, and I'm planning to add more plugins, which doens't scale up there...

sindresorhus:
Actually, nvm, it seems the config is set by default in the plugin.

kevva:
>  ... it seems the config is set by default in the plugin.

Yup :).

PR #42, opened 2015-10-22
Shareable configs

vdemedes:
This PR allows a use of eslint's `extends` property to use [shareable configs](http://eslint.org/docs/developer-guide/shareable-configs.html).

Via CLI:

```
$ xo --extend sindre
$ xo --extend eslint-config-sindre
```

Via package.json:

```js
{
  "xo": {
    "extends": "sindre"
  }
}
```

Can also be an array of configs. If `eslint-config-` prefix is missing, it will be automatically added.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -21,6 +21,7 @@ var cli = meow({
 		'  --space         Use space indent instead of tabs  [Default: 2]',
 		'  --no-semicolon  Prevent use of semicolons',
 		'  --plugin        Include third-party plugins  [Can be set multiple times]',
+		'  --extend        Extend defaults with a custom config  [Can be set multiple times]',
 		'',
 		'Examples',
 		'  $ xo',
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ var objectAssign = require('object-assign');
 var arrify = require('arrify');
 var pkgConf = require('pkg-conf');
 var deepAssign = require('deep-assign');
+var resolveFrom = require('resolve-from');
 
 var DEFAULT_IGNORE = [
 	'node_modules/**',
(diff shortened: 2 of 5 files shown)


kevva:
Why would you use xo instead of ESLint if we're going to support this?

vdemedes:
Real world example: I have some customizations over default xo rules, but I do not want to update my `xo.rules` in each package.json in all my projects. So, the solution, is to have my custom rules in `eslint-config-vdemedes` module.

This PR allows using eslint's `extends` option to be able to pull those custom rules and override xo's default ones.

sindresorhus:
I was unsure about this too, but XO already supports `rules`, so why not.

vdemedes:
It's a standard eslint option, just like any others, that `xo` exposes, so no harm done :)

sindresorhus:
Can you add a test? Just use https://github.com/sindresorhus/eslint-config-xo-react/ and steal the test logic from https://github.com/sindresorhus/eslint-config-xo-react/blob/13755dc78f7938e6da719e8991aea4b06e82cdda/test/test.js#L20-L21

cli.js
```diff
@@ -21,6 +21,7 @@ var cli = meow({
 		'  --space         Use space indent instead of tabs  [Default: 2]',
 		'  --no-semicolon  Prevent use of semicolons',
 		'  --plugin        Include third-party plugins  [Can be set multiple times]',
+		'  --extend        Extend defaults with a custom config',
```
sindresorhus: Include `[Can be set multiple times]`

vdemedes:
Done ;)

index.js
```diff
@@ -38,7 +39,9 @@ function handleOpts(opts) {
 	opts.ignores = opts.ignores || opts.ignore;
 	opts.plugins = opts.plugins || opts.plugin;
 	opts.rules = opts.rules || opts.rule;
+	opts.extends = opts.extends || opts.extend;
 
+	opts.extends = opts.extends && !Array.isArray(opts.extends) ? [opts.extends] : opts.extends;
```
sindresorhus:
```js
opts.extends = arrify(opts.extends);
```

index.js
```diff
@@ -66,6 +69,21 @@ function handleOpts(opts) {
 		opts._config.baseConfig = 'xo/esnext';
 	}
 
+	if (opts.extends) {
```
sindresorhus:
```js
if (opts.extends.length > 0) {
```

vdemedes:
Updated

sindresorhus:
Thanks!

vdemedes:
Thank you!

PR #47, opened 2015-11-13
Bump no-use-extend-native to 0.3.1

dustinspecker:
I updated this plugin in a local version of XO and ran XO on AVA and Pageres without any issues/false positives. This plugin is also being used on all of my repos. *Hopefully*, there aren't any more false positives.

I didn't re-enable the plugin - figured that's up to the XO team.


sindresorhus:
The plugin is now enabled and a new version of XO is published ;)

PR #53, opened 2015-12-03
Fixes the tests to use `test.cb` function.

brekk:
When trying to test locally I saw that the tests didn't run; updated them to use the `test.cb` asynchronous pattern.


sindresorhus:
All those tests are synchronous, so the right fix is to just remove `t.end()` instead of using `.vb` ;)

brekk:
@sindresorhus Ah, my mistake. I can create a diff. PR with that change.

PR #55, opened 2015-12-13
Open files with issues in an editor

### LINKED ISSUE #38: Open files with problems
While not directly related to xo, is there a way to open all files with problems found by xo,

`xo app/**/*.js | subl *`

… or something like that?

vdemedes:
This PR fixes #38.

When `xo` is invoked with `--open` option, it automatically opens files with issues in a user's preferred editor (by checking `$EDITOR` environment variable):

```bash
$ xo --open
```

If `$EDITOR` is empty:

screen shot 2015-12-13 at 9 45 26 am

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,7 +2,9 @@
 'use strict';
 var updateNotifier = require('update-notifier');
 var getStdin = require('get-stdin');
+var spawn = require('child_process').spawn;
 var meow = require('meow');
+var path = require('path');
 var xo = require('./');
 
 var cli = meow({
@@ -23,6 +25,7 @@ var cli = meow({
 		'  --no-semicolon  Prevent use of semicolons',
 		'  --plugin        Include third-party plugins  [Can be set multiple times]',
 		'  --extend        Extend defaults with a custom config  [Can be set multiple times]',
+		'  --open          Open files with issues in your editor',
 		'',
 		'Examples',
 		'  $ xo',
@@ -44,7 +47,8 @@ var cli = meow({
 		'init',
 		'compact',
 		'stdin',
-		'fix'
+		'fix',
+		'open'
 	]
 });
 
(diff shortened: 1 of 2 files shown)


cli.js
```diff
@@ -23,6 +24,7 @@ var cli = meow({
 		'  --no-semicolon  Prevent use of semicolons',
 		'  --plugin        Include third-party plugins  [Can be set multiple times]',
 		'  --extend        Extend defaults with a custom config  [Can be set multiple times]',
+		'  --open          Open files with issues in an editor',
```
sindresorhus: `an` => `your`

cli.js
```diff
@@ -58,6 +61,34 @@ function log(report) {
...
+		console.log();
+		console.log('`open` option was used, but your $EDITOR environment variable is empty.');
+		console.log('Fix it by setting path to your editor of choice in ~/.bashrc or ~/.zshrc:');
+		console.log();
+		console.log('    export EDITOR=atom');
+		console.log();
+		return;
+	}
+
+	spawn(editor, files).unref();
```
sindresorhus: Maybe use the `detached: true` option and `stdio: 'ignore'` just to make sure it doesn't impact XO finishing any way?

cli.js
```diff
@@ -58,6 +61,34 @@ function log(report) {
...
+	if (!editor) {
+		console.log();
+		console.log('`open` option was used, but your $EDITOR environment variable is empty.');
+		console.log('Fix it by setting path to your editor of choice in ~/.bashrc or ~/.zshrc:');
+		console.log();
+		console.log('    export EDITOR=atom');
+		console.log();
+		return;
+	}
+
```
sindresorhus: Sublime supports notation for going directly to the line and column. Would you be willing to add support for that if Sublime is the $EDITOR? (Other editors might support it too).

```
> subl --help

Filenames may be given a :line or :line:column suffix to open at a specific
location.
```

vdemedes: Was thinking about it too, cool suggestion, will implement!

P.S. Atom accepts :line too.

cli.js
vdemedes: Will add

vdemedes:
@sindresorhus PR updated according to suggestions.

cli.js
```diff
@@ -58,6 +62,44 @@ function log(report) {
...
+	var executableName = editor.split(path.sep).pop();
+
+	var files = report.results
+		.filter(function (file) {
+			return file.errorCount > 0;
+		})
+		.map(function (file) {
+			// Sublime Text and Atom support opening file at exact position
+			if (['subl', 'atom'].indexOf(executableName) >= 0) {
+				return file.filePath + ':' + file.messages[0].line;
```
sindresorhus: Can you add `column` for Sublime?

vdemedes: Done.

sindresorhus:
Perfect! This is so useful :)

cli.js
jamestalmage: Webstorm supports it as well. Default command is `wstorm`, I usually use `ws` though.

screenshot 2015-12-13 17 40 38

jamestalmage: webstorm chokes on `file:line:column`, though.

sindresorhus: @jamestalmage It will just work if you set `export $EDITOR=wstorm`. Or did you mean something else?

jamestalmage: `wstorm:line` works,
`wstorm:line:column` does not.

`xo --open` only provides line:column for sublime and atom (and it would not work just to add `wstorm` to the array).

jamestalmage: I am wondering if there is a better way to signal what syntax your editor supports.

sindresorhus: @jamestalmage Interested in doing a PR adding an exception for WebStorm (which only uses :line)?

sindresorhus: Could add support for specifying the format in `--open`, but I think that's overcomplicating it. Easier, at least for now, to just hard code what editors supports. We can easily cover the most popular ones.

jamestalmage: yeah, sure. 

I am just wondering how far you want to take supporting various editors (not that the list is likely to be super long). 

Also wondering if there is an easier way.

Qix-: Mentioned this on another PR but this could go into its own package. I could see other use cases for it.

edouard-lopez:
@jamestalmage could you provide a working command-line example using `file:line:column`

PR #57, opened 2015-12-14
Prefer local install of XO

jamestalmage:


package.json
```diff
@@ -71,9 +71,11 @@
   },
   "devDependencies": {
     "ava": "*",
+    "debug": "^2.2.0",
     "eslint-config-xo-react": "^0.3.0",
     "eslint-plugin-react": "^3.5.1",
     "execa": "^0.1.1",
+    "resolve-cwd": "^1.0.0",
```
sindresorhus: These should be in dependencies, not dev dependencies.

jamestalmage:
@sindresorhus - updated per your request

sindresorhus:
:Thanks :)

PR #58, opened 2015-12-14
Support overridable configs

### LINKED ISSUE #22: Multiple config files for subdirectories
i have a js/ folder where i want to lint with

"esnext": true,
"envs": [
  "browser"
]

and then i have my gulpfile folder where i want to lint with 

"esnext": false,
"envs": []

this works with .jshint files because i can have multiple of them in the different folders. but with xo the configuration is stored in the package.json which i tried duplicating with different rules into the different folders but it didn't work.

so i'm wondering if theres a solution already? seems like a common case.

jamestalmage:
Fixes #22.

This allows for overridable configs based on glob patterns.

Usage is as follows:

```json
{
  "xo": {
    "esnext": false,
    "overrides": [
      {
        "files": "test/*.js",
        "esnext": true
      },
      {
         "files": "test/foo.js",
         "space": 3
      }
    ]
  }
}
```

The base config is:

```json
{
  "esnext": false
}
```

For everything in the `test` directory the config is:

```json
{
  "esnext": true
}
```

With the exception of `test/foo.js` which has the following config:

```json
{
  "esnext": true,
  "space": 3
}
```

It is important to note that overrides are cumulative, and processed in the order specified in the array.
If two overrides match the same file, the latter will takes precedence over the former.

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,26 +36,28 @@ exports.lintFiles = function (patterns, opts) {
 
 		var grouped = optionsManager.groupConfigs(paths, opts, overrides);
 
-		return Promise.all(grouped.map(function (data) {
+		return mergeReports(grouped.map(function (data) {
 			return runEslint(data.paths, data.opts);
-		})).then(function (reports) {
-			// merge multiple reports into a single report
-			var results = [];
-			var errorCount = 0;
-			var warningCount = 0;
-			reports.forEach(function (report) {
-				results = results.concat(report.results);
-				errorCount += report.errorCount;
-				warningCount += report.warningCount;
-			});
-			return {
-				errorCount: errorCount,
-				warningCount: warningCount,
-				results: results
-			};
-		});
+		}));
 	});
 };
+
+function mergeReports(reports) {
+	// merge multiple reports into a single report
+	var results = [];
+	var errorCount = 0;
+	var warningCount = 0;
+	reports.forEach(function (report) {
+		results = results.concat(report.results);
+		errorCount += report.errorCount;
+		warningCount += report.warningCount;
+	});
+	return {
+		errorCount: errorCount,
+		warningCount: warningCount,
+		results: results
+	};
+}
 
 function runEslint(paths, opts) {
 	var config = optionsManager.buildConfig(opts);
--- a/index.js
+++ b/index.js
@@ -2,17 +2,11 @@
 var path = require('path');
 var eslint = require('eslint');
 var globby = require('globby');
-var optsHandler = require('./opts');
-
-function optsToConfig(opts) {
-	opts = optsHandler.mergeWithPkgConf(opts);
-	opts = optsHandler.normalizeOpts(opts);
-	opts.ignores = opts.ignore = optsHandler.DEFAULT_IGNORE.concat(opts.ignores || []);
-	return optsHandler.buildConfig(opts);
-}
+var optionsManager = require('./options-manager');
 
 exports.lintText = function (str, opts) {
-	opts = optsToConfig(opts);
+	opts = optionsManager.preprocess(opts);
+	opts = optionsManager.buildConfig(opts);
 
 	var engine = new eslint.CLIEngine(opts);
 
@@ -20,7 +14,7 @@ exports.lintText = function (str, opts) {
 };
 
 exports.lintFiles = function (patterns, opts) {
-	opts = optsToConfig(opts);
+	opts = optionsManager.preprocess(opts);
 
 	if (patterns.length === 0) {
 		patterns = '**/*.{js,jsx}';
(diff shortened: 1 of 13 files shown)


jamestalmage:
We had discussed using object hashes for overrides:

```js
"overrides": {
  "**/foo.js": { "esnext": false } 
}
```

I decided against doing it that way, because it gives you no control over merge order if two patterns match the same file.

jamestalmage:
There is still debate over exactly what to name some of these options. Once that is settled, I will update this PR (including docs).

SamVerschueren:
The overrides functionality is really nice! Is it possible to provide a concrete example on the benefit of the array versus the object hashes? It seems like it just makes everything more complex because the order in the array is important. It's because of merge conflicts when more globbing patterns matches the same file, right? But why not apply the most specific rules for that file?

```json
"overrides": {
  "test/test.js": {esnext: true},
  "test/*.js": { "esnext": false } 
}
```

Because `test/test.js` is more specific, the `esnext` rule should be `true` for the file `test/test.js`. Independent of the order of the keys.

Just wanted some more information regarding this :).

jamestalmage:
> It's because of merge conflicts when more globbing patterns matches the same file, right?

Yes. The array syntax is necessary to support multiple glob patterns anyways:

```js
"overrides": [
  {
    "files": ["test/*.js", "!test/foo.js"]
  }
]
```

So I think the array syntax needs to be a part of the final solution no matter what. It might be possible to allow both array and object.

To be honest, the array solution was just easier to implement.

> But why not apply the most specific rules for that file?

I am not sure how to reliably determine "the most specific match".  Which of the following more specifically matches `test/foo.js`?

- `**/foo.js`
- `test/*oo.js`

Also, how do we then merge settings. Least specific to most specific? In that case the following is confusing:

```js
"overrides": {
  "test/test.js": {"esnext": true},
  "test/*.js": { "esnext": false, spaces:3} 
}
```

If we merge every match, in order of specificity., the settings for `test/test.js` are obtained by merging the least specific (`{esnext: false, spaces: 3}`), followed by the most specific (`{esnext:true}`). But that's opposite the order I have them listed above. That is confusing.

I think if we allow the object pattern we should only apply a single "most specific" match (I am still not sure how best to determine that).

SamVerschueren:
> The array syntax is necessary to support multiple glob patterns anyways

Good point

> I am not sure how to reliably determine "the most specific match"

Not sure either

> But that's opposite the order I have them listed above. That is confusing.

I don't think that's confusing. But maybe that's just me.

In general, whatever gets implemented, as long as it's clearly documented, we're good :).

jamestalmage:
Yeah, I fully intend to document before it gets merged. Still looking for feedback at this point.

options-manager.js
```diff
@@ -0,0 +1,179 @@
...
+		config.rules['semi-spacing'] = [2, {before: false, after: true}];
+	}
+
+	if (opts.esnext) {
+		config.baseConfig.extends = 'xo/esnext';
+	} else {
+		// always use the Babel parser so it won't throw
+		// on esnext features in normal mode
+		config.parser = 'babel-eslint';
+		config.plugins = ['babel'];
```
jamestalmage: Also, I noticed that if `opts.esnext` is not `true`, we overwrite any plugins they might have specified. Should we just be appending/prepending babel to the list?

sindresorhus: Oh, noes. That's why https://github.com/sindresorhus/xo/commit/90c95eda4851aac058a6b846b2cd63a0263d1f33#commitcomment-14948953 Good eye! :)

options-manager.js
```diff
@@ -0,0 +1,179 @@
...
+
+// Creates grouped sets of merged options together with the paths they apply to.
+function groupConfigs(paths, baseOptions, overrides) {
+	var map = {};
+	var arr = [];
+
+	paths.forEach(function (x) {
+		var data = findApplicableOverrides(x, overrides);
+		if (!map[data.hash]) {
+			var mergedOpts = deepAssign.apply(null, [{}, baseOptions].concat(data.applicable));
```
jamestalmage: This is where the merging of every applicable "overrides" happens. Is the `deepAssign` logic sufficient here?

options-manager.js
jamestalmage: @sindresorhus You want me to fix it here or a separate PR from `master`?

sindresorhus: @jamestalmage Master. I need to do a patch release. :)

jamestalmage:
@sindresorhus @kevva @vdemedes - Do you guys have any additional feedback? If I get this rebased is it good to merge?

vdemedes:
Good work, looking forward to using this!

options-manager.js
```diff
@@ -0,0 +1,179 @@
+'use strict';
+var path = require('path');
+var arrify = require('arrify');
+var pkgConf = require('pkg-conf');
+var deepAssign = require('deep-assign');
+var resolveFrom = require('resolve-from');
+var objectAssign = require('object-assign');
+var homeOrTmp = require('home-or-tmp');
+var mutlimatch = require('multimatch');
```
sindresorhus: `mutlimatch` => `multimatch`

index.js
```diff
@@ -130,12 +27,42 @@ exports.lintFiles = function (patterns, opts) {
...
+
+		var overrides = opts.overrides;
+		delete opts.overrides;
+
+		var grouped = optionsManager.groupConfigs(paths, opts, overrides);
+
+		return Promise.all(grouped.map(function (data) {
+			return runEslint(data.paths, data.opts);
+		})).then(function (reports) {
+			// merge multiple reports into a single report
```
sindresorhus: The below could be in a separate function called something like `mergeReports`.

index.js
```diff
@@ -130,12 +27,42 @@ exports.lintFiles = function (patterns, opts) {
...
+		if (!(opts.overrides && opts.overrides.length)) {
+			return runEslint(paths, opts);
+		}
+
+		var overrides = opts.overrides;
+		delete opts.overrides;
+
+		var grouped = optionsManager.groupConfigs(paths, opts, overrides);
+
+		return Promise.all(grouped.map(function (data) {
```
sindresorhus: Why use `Promise.all` here? `runEslint` and `engine.executeOnFiles` are sync functions.

options-manager.js
sindresorhus: Yes

sindresorhus:
Sorry about the delay. I've just had too many PRs to review. Took a thorough look now and it generally looks fantastic! Also thanks for refactoring the code into a more readable and maintainable state 👌.

I added a few minor inline comments, but it's pretty close to be mergeable. Currently, all the individual pieces are tested, but could use a real world integration test in `test.js` with a fixture and a real package.json, just to ensure all the "glue" works correctly. Also needs docs, but I would be happy to write that if you don't want / have time.

It's 🎄so no rush on this :)

jamestalmage:
@sindresorhus - this should be all set.

I added an undocumented `--no-local` flag to disable the local check. Otherwise we were getting our old version for all the tests in `test/cli.js`

jamestalmage:
@sindresorhus - Are we good to merge here? I would like to avoid rebasing again. 

Once merged I suggest we let it live on master and test it against a few projects before release. It was a significant refactor followed by a problematic rebase. I'm fairly confident I got it right, but better to practice caution.

sindresorhus:
Yes, landed :D I'm super happy about how this turned out.

unicorn

sindresorhus:
Gonna let it live on master for a few days while I test it on some of my projects.

dvdln:
...any chance this will be released anytime soon?

sindresorhus:
@dvdln When this issue is resolved: https://github.com/sindresorhus/xo/issues/71

PR #59, opened 2015-12-15
Don't destroy plugins.

jamestalmage:
If the `esnext` option is `false`, we currently blow away existing
plugins, replacing it with only the `babel` plugin. This prepends it.

This does not come with a test, as option handling is currently pretty difficult.
We can add the appropriate test for this there.

Reference:
  https://github.com/sindresorhus/xo/pull/58#discussion_r47583285


sindresorhus:
Thank you @jamestalmage for noticing this :D

PR #62, opened 2015-12-19
ignore all node_module and bower_components directories

brandon93s:


index.js
```diff
@@ -11,6 +11,7 @@ var homeOrTmp = require('home-or-tmp');
 
 var DEFAULT_IGNORE = [
 	'node_modules/**',
+	'**/node_modules/**',
```
sindresorhus: I think this makes the rule above moot.

index.js
```diff
@@ -11,6 +11,7 @@ var homeOrTmp = require('home-or-tmp');
...
 	'bower_components/**',
```
sindresorhus: Should also be done here.

brandon93s:
Updated globs and squashed commits.

index.js
Qix-: If it's a glob pattern, try `node_modules/**/*`.

sindresorhus: @Qix- That wouldn't match `foo/node_modules/`

Qix-: why wouldn't it?

brandon93s: Trailing `/**` matches all directories and subdirectories. `/**/*` matches all files in all directories and subdirectories. 

In this case, they're the same. The prior is simpler. The former would be to match a certain file type: 
`/**/*.js`

Qix-: Gotcha.

sindresorhus:
Looks good. Thanks :)

PR #63, opened 2015-12-29
Merge custom rules after defaults are determined

tcoenraad:
Custom rules like 
```js
"babel/object-curly-spacing": [2, "always"]
```
are no longer ignored, as they are merged after defaults are determined.

### DIFF
--- a/options-manager.js
+++ b/options-manager.js
@@ -71,13 +71,9 @@ function buildConfig(opts) {
 		envs: opts.envs,
 		globals: opts.globals,
 		plugins: DEFAULT_PLUGINS.concat(opts.plugins || []),
-		rules: opts.rules,
+		rules: {},
 		fix: opts.fix
 	});
-
-	if (!config.rules) {
-		config.rules = {};
-	}
 
 	if (opts.space) {
 		var spaces = typeof opts.space === 'number' ? opts.space : 2;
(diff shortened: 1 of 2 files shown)


jamestalmage:
LGTM. Can you squash?

options-manager.js
```diff
@@ -102,6 +98,14 @@ function buildConfig(opts) {
 		config.rules['babel/object-curly-spacing'] = [2, 'never'];
 	}
 
+	if (opts.rules) {
+		for (var attr in opts.rules) {
```
jamestalmage: Object.keys(opts.rules).forEach ?

sindresorhus: :+1:

options-manager.js
```diff
@@ -102,6 +98,14 @@ function buildConfig(opts) {
...
+			if (opts.rules.hasOwnProperty(attr)) {
+				config.rules[attr] = opts.rules[attr];
+			}
```
sindresorhus: Why not just `objectAssign`?

tcoenraad: Because of this https://github.com/sindresorhus/xo/commit/bdfccc6db0fd3c1f3fae81128ce3dd7a9074cb6c

tcoenraad:
Changes are now squashed.

options-manager.js
tcoenraad: Thanks, I used this approach eventually.

options-manager.js
sindresorhus: You can just use the objectAssign ponyfill. We already depend on it https://github.com/sindresorhus/xo/blob/4b44968e18666bf22f742de9af12f4b61315cd8e/options-manager.js#L7

tcoenraad: I like the word ponyfill. I altered my commit accordingly.

sindresorhus:
Thank you! I'll do another release this weekend :)

Issue #65, opened 2016-01-03
A way to override the default ignores


jamestalmage:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/xojs/xo/issues/65)

Right now it's not possible to lint anything in a `fixtures` directory, even if you want to. There should be a way to enable that. 

I tried this: 

``` js
{
  "xo": {
    "ignore": [
      "!{test/,}fixture{s,}/**"
    ]
  }
}
```

But no luck

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#425 Find XO config based on linted file path](https://issuehunt.io/r/xojs/xo/pull/425)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/xojs/xo/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

sindresorhus:
I would have thought that would work, and it definitely should. XO comes with good defaults, but everything should be overridable when needed.

jamestalmage:
If anybody wants to take a crack at this #66 has a test prewritten for you!

jahson:
@jamestalmage Can you please tell me if ignore in XO config should totally override the default ignore? In current code it's [appended to](https://github.com/sindresorhus/xo/blob/b6199c0f93312f5f5a615b3d9d3eaa8fb5a2d65f/options-manager.js#L194) `DEFAULT_IGNORE` and then [passed to](https://github.com/sindresorhus/xo/blob/b6199c0f93312f5f5a615b3d9d3eaa8fb5a2d65f/index.js#L23) `globby` which in turn [passes it to](https://github.com/sindresorhus/globby/blob/51de1a7278bd32dcb8cde2978f42181609d43f93/index.js#L51) `glob`.

jamestalmage:
> Can you please tell me if ignore in XO config should totally override the default ignore? 

[...]

PR #66, opened 2016-01-03
Add a failing test for #65.

jamestalmage:
See #65.

I haven't figured out the exact cause of the problem, but a fix for #65 should make this fixture pass.

I've deployed to @sindresorhus's repo, so anyone should be able to just `checkout add-ignores-test-fixture` if they want to take a crack at solving the problem.


jamestalmage:
5010bd7 is probably worthwhile on it's own.

sindresorhus:
:+1:

test/fixtures/ignores/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "application-name",
+  "version": "0.0.1",
+  "xo": {
+    "ignore": [
+      "{test/,}fixture{s,}/**"
```
sindresorhus: @jamestalmage Looking at this again, should it be prefixed with `!`?

jamestalmage: yes.

PR #81, opened 2016-02-15
add fly-xo to plugins section

lukeed:


sindresorhus:
> The xo task must be last function within your task.

Why?

lukeed:
Yeah, it follows suit to how all the other Fly linters are set up. It has to do with the plugin structure, and is actually something that will be reworked in Fly 1.0.

Thanks for merge :bow:

Issue #71, opened 2016-01-08
Unable to customize certain rules via shareable config


vdemedes:
Using latest xo (master branch), I'm unable to customize `babel/object-curly-spacing` rule via shareable config.

package.json:

```js
{
  "extends": "my-config"
}
```

my-config:

```js
{
  "rules": {
    "babel/object-curly-spacing": 0
  }
}
```

file.js:

```js
{ key: 'value' }
```

Given these files, xo throws error:

<img width="570" alt="screen shot 2016-01-08 at 10 18 49 pm" src="https://cloud.githubusercontent.com/assets/697676/12208509/d5c9078e-b655-11e5-9373-0ff60e116789.png">

If I override `babel/object-curly-spacing` in package.json, it works as expected and I no longer get errors:

```js
{
  "rules": {
    "babel/object-curly-spacing": 0
  }
}
```

Related: https://github.com/sindresorhus/xo/blob/master/options-manager.js#L98

sindresorhus:
> Related: https://github.com/sindresorhus/xo/blob/2d26d79ea81f893fc7b1ff9e586c1e74ad28c830/options-manager.js#L98

Hmm. We just set the default value there. It's later extended by your custom config.

vdemedes:
> It's later extended by your custom config.

Seems like it does not for some reason. Needs investigation.

kevva:
Those rules gets set in the `root` of the options sent in to `eslint` and will therefore take precedence over any configs. Those are only overridable through setting `rules` in the `package.json`, or wherever your config lives.

vdemedes:
@kevva is right.

Why does xo forces these default rules here (https://github.com/sindresorhus/xo/blob/2d26d79ea81f893fc7b1ff9e586c1e74ad28c830/options-manager.js#L98) ?

jamestalmage:
What would be the consequence of moving those defaults into another base config that we conditionally add to `extends`, instead of force setting them like that?

kevva:
I don't think there are any consequences other than creating a config for it may seem unnecessary but in this case it might be needed.

vdemedes:
Why are those rules not in the default xo config in the first place?

sindresorhus:
> Why does xo forces these default rules here (https://github.com/sindresorhus/xo/blob/2d26d79ea81f893fc7b1ff9e586c1e74ad28c830/options-manager.js#L98) ?

[...]

PR #83, opened 2016-03-02
Demo broken shareable configs

vdemedes:
PR to demonstrate https://github.com/sindresorhus/xo/issues/71#issuecomment-191247202. The test project is in `test/fixtures/project`.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -31,3 +31,8 @@ test.skip('ignores fixture', async t => {
 	const cwd = path.join(__dirname, 'fixtures/ignores');
 	t.throws(execa('../../../cli.js', ['--no-local'], {cwd}));
 });
+
+test('supports shareable config', async () => {
+	const cwd = path.join(__dirname, 'fixtures/project');
+	await execa('../../../cli.js', ['--no-local'], {cwd});
+});
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+!test/fixtures/project/node_modules
--- a/test/fixtures/project/file.js
+++ b/test/fixtures/project/file.js
@@ -1,0 +1,2 @@
+var obj = { a: 1 };
+console.log(obj.a);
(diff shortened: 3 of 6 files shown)


sindresorhus:
Cool. Thanks for doing this. Could you make it something I could merge? So that tests can just fail until it's fixed.

vdemedes:
You mean to `.skip()` the failing test?

package.json
```diff
@@ -94,5 +94,8 @@
     "ignores": [
       "cli.js"
     ]
+  },
+  "ava": {
+    "files": "test/cli.js"
```
sindresorhus: I mean, remove this.

test/cli.js
```diff
@@ -31,3 +31,9 @@ test.skip('ignores fixture', async t => {
 	const cwd = path.join(__dirname, 'fixtures/ignores');
 	t.throws(execa('../../../cli.js', ['--no-local'], {cwd}));
 });
+
+test.only('broken shareable config', async () => {
```
sindresorhus: And make this just `test()`

sindresorhus: And rename it to something like:

> supports shareable extended config

vdemedes:
Done!

test/cli.js
```diff
@@ -31,3 +31,9 @@ test.skip('ignores fixture', async t => {
 	const cwd = path.join(__dirname, 'fixtures/ignores');
 	t.throws(execa('../../../cli.js', ['--no-local'], {cwd}));
 });
+
+test('supports shareable config', async () => {
+	await execa('rm', ['-rf', '../node_modules/xo'], {cwd: __dirname});
```
sindresorhus: Sorry, didn't notice this. Why is this needed?

vdemedes: Oh yes, I guess this is another, separate issue. https://github.com/sindresorhus/xo/blob/master/package.json#L91 causes `cli.js` to stop execution and rely on `node_modules/xo` rather than version in `master` branch. So when you're testing `cli.js`, you are not actually testing it, because it spawns `cli.js` from `node_modules/xo`.

sindresorhus: Hmm, that's what the `--no-local` flag is supposed to prevent, though.

vdemedes: Oh ok, I'll change it.

vdemedes: Done ;) Totally my ~~overshit~~ oversight.

PR #94, opened 2016-03-28
Make `--spaces` apply to JSX (i.e., override `eslint-config-xo-react`).

sterpe:
Hello @sindresorhus,

I have to work on this JSX project that insists on 4 space tabs, but discovered that `xo`'s `--space` flag wasn't overriding the JSX indentation set by `eslint-config-xo-react`.

This PR overrides the JSX spacing rules when the `--space` flag is used.

Thanks!


sterpe:
I guess it's simple enough to resolve with an overriding shareable config of our own.  I'll leave it to your judgement whether this has any value as an intuitive feature.

PR #102, opened 2016-04-09
[WIP]: fix option handling.

jamestalmage:
I haven't explored why I was doing it the way I was, but this passes all the tests.


sindresorhus:
You can now remove the workaround in https://github.com/sindresorhus/xo/blob/5e5c4efa1a1a561e12a26b73c71410d69b74407b/index.js#L8

PR #113, opened 2016-05-06
Give stdout time to flush.

jamestalmage:
I do not know why, but I'm suddenly seeing the process exit before the entire report prints.

This fixes the problem. I really wish there was a better way. This just feels wrong.

What I was seeing:

screenshot 2016-05-05 21 11 05

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -83,7 +83,11 @@ function log(report) {
 	var reporter = opts.reporter ? xo.getFormatter(opts.reporter) : formatterPretty;
 
 	process.stdout.write(reporter(report.results));
-	process.exit(report.errorCount === 0 ? 0 : 1);
+
+	setTimeout(function () {
+		// give stdout time to flush.
+		process.exit(report.errorCount === 0 ? 0 : 1);
+	}, 100);
 }
 
 function open(report) {
--- a/cli.js
+++ b/cli.js
@@ -22,6 +22,7 @@ var meow = require('meow');
 var path = require('path');
 var formatterPretty = require('eslint-formatter-pretty');
 var xo = require('./');
+var exit = require('exit');
 
 var cli = meow({
 	help: [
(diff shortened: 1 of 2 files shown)


Qix-:
Not crazy about this approach. We should try the `exit` module until the root problem is fixed. A time based approach is definitely not the solution here, even temporarily.

jamestalmage:
Nice! Didn't know about the `exit` module. AVA has been plagued by this, and we resorted to timers there. I'm going to need to fix that!

Qix-:
@jamestalmage the failing test's assertion is this:

```
1. cli » fix option
  AssertionError: 'console.log(0)' === 'console.log(0);'
    _callee$ (cli.js:12:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)
```

jamestalmage:
Yes, but why?

That appears to be a test for the `--fix` option. Can't understand why it's being affected (and only on Windows).

jamestalmage:
Test failures are related to https://github.com/IndigoUnited/node-cross-spawn-async/pull/19

karinepires:
I'm having the same error and I've tried to correct it before finding this PR.
I did with the following code (as indicated on [node docs](https://nodejs.org/api/process.html#process_process_stdout)), is there any reason to not use it?
`process.exitCode = report.errorCount === 0 ? 0 : 1;`

sindresorhus:
@karinepires This is being fixed in Node.js itself. Just don't use Node.js 6 for now.

sindresorhus:
Node.js 6.2.1:

> tty: Default to blocking mode for stdio on OS X. A bug fix in libuv 1.9.0, introduced in Node.js v6.0.0, exposed problems with Node's use of non-blocking stdio, particularly on OS X which has a small output buffer. This change should fix CLI applications that have been having problems with output since Node.js v6.0.0 on OS X. The core team is continuing to address stdio concerns that exist across supported platforms and progress can be tracked at https://github.com/nodejs/node/issues/6980. (Jeremiah Senkpiel) #6895 - https://nodejs.org/en/blog/release/v6.2.1/

Qix-:
Nice.

PR #142, opened 2016-09-14
Add `settings` and `parser` options

### LINKED ISSUE #138: No way to pass `settings`?
The eslint-import checker is nice but it needs to be told about my webpack config. However, I don't see how to provide the `settings` key to eslint.

wmertens:
Closes #138.

### DIFF
--- a/test/options-manager.js
+++ b/test/options-manager.js
@@ -76,6 +76,12 @@ test('buildConfig: rules', t => {
 	t.deepEqual(config.rules, rules);
 });
 
+test('buildConfig: parser', t => {
+	const parser = 'babel-eslint';
+	const config = manager.buildConfig({parser});
+	t.deepEqual(config.baseConfig.parser, parser);
+});
+
 test('buildConfig: settings', t => {
 	const settings = {'import/resolver': 'webpack'};
 	const config = manager.buildConfig({settings});
--- a/test/options-manager.js
+++ b/test/options-manager.js
@@ -15,6 +15,7 @@ test('normalizeOpts: makes all the opts plural and arrays', t => {
 		ignore: 'test.js',
 		plugin: 'my-plugin',
 		rule: {'my-rule': 'foo'},
+		setting: {'my-rule': 'bar'},
 		extend: 'foo',
 		extension: 'html'
 	});
@@ -25,6 +26,7 @@ test('normalizeOpts: makes all the opts plural and arrays', t => {
 		ignores: ['test.js'],
 		plugins: ['my-plugin'],
 		rules: {'my-rule': 'foo'},
+		settings: {'my-rule': 'bar'},
 		extends: ['foo'],
 		extensions: ['html']
 	});
@@ -72,6 +74,12 @@ test('buildConfig: rules', t => {
 	const rules = {'object-curly-spacing': ['error', 'always']};
 	const config = manager.buildConfig({rules});
 	t.deepEqual(config.rules, rules);
+});
+
+test('buildConfig: settings', t => {
+	const settings = {'import/resolver': 'webpack'};
+	const config = manager.buildConfig({settings});
+	t.deepEqual(config.baseConfig.settings, settings);
 });
 
 test('findApplicableOverrides', t => {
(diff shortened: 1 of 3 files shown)


wmertens:
Weird thing though, when I use this to set the `import/resolver`, it makes it into the config that is generated for ESLint, but the import plugin only gets the rule set by the `plugins` extend.

generated config:
```
{ rules: 
   { strict: 0,
     quotes: 0,
     eqeqeq: [ 2, 'allow-null' ],
     'no-eq-null': 0,
     'operator-linebreak': [ 2, 'after', [Object] ],
     'brace-style': [ 2, '1tbs', [Object] ],
     'babel/arrow-parens': 0,
     'comma-dangle': [ 2, 'always-multiline' ],
     'react/jsx-closing-bracket-location': [ 2, 'tag-aligned' ],
     'react/prop-types': 1,
     'react/prefer-stateless-function': 1,
     'prefer-template': 0,
     semi: [ 2, 'never' ],
     'semi-spacing': [ 2, [Object] ] },
  settings: { 'import/resolver': { webpack: [Object] } },
  globals: [ '$', 'info', 'dbg', 'views', '_dict', 'dict', 'lang' ],
  envs: [ 'browser', 'mocha', 'webpack' ],
  plugins: [ 'react', 'babel' ],
  extends: [ 'xo-react' ],
  useEslintrc: false,
  cache: true,
  cacheLocation: '/Users/wmertens/.xo-cache/',
  baseConfig: 
   { extends: 
      [ 'xo/esnext',
        '/Users/wmertens/.../src/node_modules/xo/config/plugins.js',
        '/Users/wmertens/.../src/node_modules/eslint-config-xo-react/index.js' ] },
  esnext: true,
  semicolon: false,
  cwd: '/Users/wmertens/.../src',
  init: false,
  compact: false,
  stdin: false,
  fix: false,
  open: false,
  ignores: 
   [ '**/node_modules/**',
     '**/bower_components/**',
     'coverage/**',
     '{tmp,temp}/**',
     '**/*.min.js',
     '**/bundle.js',
     'fixture{-*,}.{js,jsx}',
     'fixture{s,}/**',
     '{test,tests,spec,__tests__}/fixture{s,}/**',
     'vendor/**',
     'dist/**' ] }
```

settings that the import plugin receives:
```
{ 'import/extensions': [ '.js' ] }
```

Does eslint not properly merge the settings?

wmertens:
Meep?

test/options-manager.js
```diff
@@ -73,6 +75,12 @@ test('buildConfig: rules', t => {
 	t.deepEqual(config.rules, rules);
 });
 
+test('buildConfig: settings', t => {
+	const settings = {'import/resolver': 'webpack'};
+	const config = manager.buildConfig({settings: settings});
```
sindresorhus:
```js
const config = manager.buildConfig({settings});
```

sindresorhus:
> Weird thing though, when I use this to set the import/resolver, it makes it into the config that is generated for ESLint, but the import plugin only gets the rule set by the plugins extend.

// @jfmengels in case you have any idea what could cause this.

jfmengels:
Sorry, don't have a clue :confused:

wmertens:
Done

sindresorhus:
Ok, figured it out. It's because we pass these option to the [ESLint Node.js API](http://eslint.org/docs/developer-guide/nodejs-api), not as normal config (since it's generated). The ESLint Node.js API doesn't yet have a `settings` option. @wmertens Can you open an issue on ESLint about adding that?

wmertens:
@sindresorhus it turns out you're supposed to pass the config via `baseconfig`…

sindresorhus:
@wmertens Ah ok. Makes sense. Can you update to use `baseConfig` instead? (See `extends` for example).

sindresorhus:
No pressure on this, but I plan on doing a new release in a few days, so this needs to be updated if you want it included ;)

wmertens:
Noted, will try ☺

sindresorhus:
Last chance. Tomorrow is release day ;)

wmertens:
@sindresorhus Alright, this should do the trick…

sindresorhus:
Awesome. Glad we could land this before the release. Thank you @wmertens :)

PR #147, opened 2016-09-30
Ignore paths from `.gitignore`

### LINKED ISSUE #27: Automatically ignore directories/files in .gitignore
To avoid all those pesky babel and electron build dirs.

juanj:
Fixes #27


sholladay:
I don't think it should be behind a flag. The whole point is to reduce the burden on the user. In fact, I'd personally avoid an option until we hear of a use case for disabling it.

juanj:
So will be better if the flags is to disable it?

vdemedes:
Totally agree with @sholladay, the `--use-gitignore` flag is unnecessary.

vdemedes:
@juanj Could you add a simple test, confirming it actually skips the patterns in `.gitignore`? I think it should be good to go then!

package.json
```diff
@@ -82,7 +82,8 @@
     "resolve-cwd": "^1.0.0",
     "resolve-from": "^2.0.0",
     "update-notifier": "^1.0.0",
-    "xo-init": "^0.3.0"
+    "xo-init": "^0.3.0",
+    "parse-gitignore": "^0.3.1"
```
kevva: These should be listed alphabetically.

options-manager.js
```diff
@@ -195,6 +196,10 @@ function preprocess(opts) {
 	opts = mergeWithPkgConf(opts);
 	opts = normalizeOpts(opts);
 	opts.ignores = DEFAULT_IGNORE.concat(opts.ignores || []);
+
+	const gitignore = parseGitignore('.gitignore');
+	opts.ignores = opts.ignores.concat(gitignore || []);
```
kevva: Create a function instead, `getIgnores` or something, that we then can export like we've done with the other ones.

options-manager.js
```diff
@@ -191,10 +192,20 @@ function groupConfigs(paths, baseOptions, overrides) {
 	return arr;
 }
 
+function getIgnores(opts) {
+	const gitignore = parseGitignore('.gitignore');
+
+	opts.ignores = DEFAULT_IGNORE.concat(opts.ignores || []);
+	opts.ignores = opts.ignores.concat(gitignore || []);
```
sindresorhus: No need for multiple concat statement. `Array#concat` support multiple arguments.

options-manager.js
```diff
@@ -191,10 +192,20 @@ function groupConfigs(paths, baseOptions, overrides) {
 	return arr;
 }
 
+function getIgnores(opts) {
+	const gitignore = parseGitignore('.gitignore');
```
sindresorhus: This needs to take into account the `cwd` option.

test/options-manager.js
```diff
@@ -136,3 +136,8 @@ test('groupConfigs', t => {
 		return obj;
 	}));
 });
+
+test('gitignore', t => {
+	const result = manager.getIgnores({});
+	t.not(result.ignores.indexOf('foo/**'), -1);
```
sindresorhus: Use `Array#includes`

sindresorhus:
What if there are nested `.gitignore`'s?

```
.
├── .gitignore
├── foo
│   ├── .gitignore
│   └── cake.js
└── unicorn.js
```

test/options-manager.js
juanj: It seems that method does not work in node v4.
Should we return to indexOf?

options-manager.js
```diff
@@ -193,11 +194,24 @@ function groupConfigs(paths, baseOptions, overrides) {
 }
 
 function getIgnores(opts) {
-	const gitignore = parseGitignore('.gitignore');
+	const gitignores = globby.sync('**/.gitignore');
+	let ignores = [];
 
-	opts.ignores = DEFAULT_IGNORE.concat(opts.ignores || []);
-	opts.ignores = opts.ignores.concat(gitignore || []);
+	gitignores.forEach(path => {
```
kevva: `for (const path of gitignores) {`

options-manager.js
```diff
@@ -193,11 +194,24 @@ function groupConfigs(paths, baseOptions, overrides) {
...
+		const location = path.substring(0, path.length - 10);
+		const fullPathResult = [];
 
+		result.forEach(file => {
+			file = location + file;
+			fullPathResult.push(file);
+		});
+
+		ignores = ignores.concat(fullPathResult || []);
+	});
```
kevva: Just do the following:

```js
const ignores = gitignores
    .map(path => {
        const res = parseGitignore(path);
        const location = path.substring(0, path.length - 10);

        return res.map(file => location + file);
    })
    .reduce((a, b) => a.concat(b));
```

options-manager.js
```diff
@@ -191,10 +193,33 @@ function groupConfigs(paths, baseOptions, overrides) {
 	return arr;
 }
 
+function getIgnores(opts) {
+	const gitignores = globby.sync('**/.gitignore');
+	let ignores = [];
+
+	gitignores.forEach(path => {
+		const result = parseGitignore(path);
+		const location = path.substring(0, path.length - 10);
```
kevva: Why do we do this?

options-manager.js
```diff
@@ -191,10 +193,27 @@ function groupConfigs(paths, baseOptions, overrides) {
 	return arr;
 }
 
+function getIgnores(opts) {
+	const gitignores = globby.sync('**/.gitignore');
```
sindresorhus: It should only look for `.gitignore` files in not already ignored locations. So you need to pass the existing ignores here (in the `ignore` option). Right now it also looks in for example the `node_modules` folder, which is very slow.

Should also be a test for this (API test).

options-manager.js
```diff
@@ -191,10 +193,27 @@ function groupConfigs(paths, baseOptions, overrides) {
...
+	const ignores = gitignores
+	.map(pathToGitignore => {
+		const result = parseGitignore(pathToGitignore);
+		const location = path.dirname(pathToGitignore);
```
sindresorhus: `location` => `base`
`result` => `patterns`

options-manager.js
sindresorhus: This must also take into account the XO `cwd` option.

test/cli.js
```diff
@@ -31,6 +31,17 @@ test.failing('ignores fixture', async t => {
 	t.throws(execa('../../../cli.js', ['--no-local'], {cwd}));
 });
 
+test('ignore files in .gitignore', async t => {
```
sindresorhus: Would be good if this test was duplicated as an API test too (using the `cwd` option).

sindresorhus:
The merge conflict needs to be resolved.

sindresorhus:
It also turns out `parse-gitignore` produces incorrect globs... `node_modules` in a .gitignore is equivalent to `**/node_modules/**`, but `parse-gitignore` produces `node_modules/**`. @juanj Can you open an issue on `parse-gitignore`? 

See: https://git-scm.com/docs/gitignore

sindresorhus:
@juanj Any chance you could finish this up? :)

test/cli.js
juanj: Why this should be an API test?
What are the API tests for?

sindresorhus:
Looks great now. Thanks for working on this @juanj :)

test/cli.js
sindresorhus: To make sure the `cwd` option is adhered to. In the CLI test it just uses `process.cwd()`. API tests are a lot faster to run too. Long-term, I'm going to convert many of the CLI tests to API tests.

I've merged this now, but would be nice if you could submit an API test for this.

Issue #117, opened 2016-05-21
Allowing more file extensions


t2t2:
Currently file extensions are [limited to .js and .jsx](https://github.com/sindresorhus/xo/blob/master/index.js#L27), but that limits the use to those extensions even if the pattern is set to match some other extensions that eslint can lint (either standalone or with a plugin, for example `**/*.{js,vue}` with eslint-plugin-html)

sindresorhus:
Yeah, agreed. It should only limit to .js/jsx if you don't specify a file extension in the glob.

PR #149, opened 2016-10-04
Add `extensions` option

addobandre:
Solves #117 (Allowing more file extensions)

Trying to detect whether the glob pattern restricts file extensions quickly gets out of hand. There are arrays of patterns, there can be negations, obscure regex syntax that may not be accounted for and that does limit what extensions can be used...

So, I added a new option: anyExtension (or --any-extension).
Set it to true, and the extension filter is disabled.

In my opinion, it would be better to just remove it altogether. Most people are probably just using the default expression and the ignores option, so impact should be minimal. A FAQ entry would be fine. Maybe this should happen for 1.0?

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -42,7 +42,7 @@ const cli = meow(`
 	  --extend        Extend defaults with a custom config  [Can be set multiple times]
 	  --open          Open files with issues in your editor
 	  --quiet         Show only errors and no warnings
-	  --extension     Additional extensions to lint [Can be set multiple times]
+	  --extension     Additional extension to lint [Can be set multiple times]
 
 	Examples
 	  $ xo
--- a/cli.js
+++ b/cli.js
@@ -28,20 +28,21 @@ const cli = meow(`
 	  $ xo [<file|glob> ...]
 
 	Options
-	  --init          Add XO to your project
-	  --fix           Automagically fix issues
-	  --reporter      Reporter to use
-	  --stdin         Validate code from stdin
-	  --esnext        Enforce ES2015+ rules
-	  --env           Environment preset  [Can be set multiple times]
-	  --global        Global variable  [Can be set multiple times]
-	  --ignore        Additional paths to ignore  [Can be set multiple times]
-	  --space         Use space indent instead of tabs  [Default: 2]
-	  --no-semicolon  Prevent use of semicolons
-	  --plugin        Include third-party plugins  [Can be set multiple times]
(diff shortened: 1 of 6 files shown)


sindresorhus:
I think the correct solution would be to introduce an `extensions` option where users can specify additional extensions to include.

addobandre:
Changed it to the `extensions` option. Should I change the default pattern to `**/*`? That way, someone could just enable more extensions and not need to change anything else. It shouldn't break anything since the filter has the usual .js/.jsx by default.

cli.js
```diff
@@ -42,6 +42,7 @@ const cli = meow(`
 	  --extend        Extend defaults with a custom config  [Can be set multiple times]
 	  --open          Open files with issues in your editor
 	  --quiet         Show only errors and no warnings
+	  --extension     Additional extensions to lint [Can be set multiple times]
```
sindresorhus: extensions -> extension

index.js
```diff
@@ -32,10 +32,12 @@ exports.lintFiles = (patterns, opts) => {
...
 	return globby(patterns, {ignore: opts.ignores}).then(paths => {
-		// when users are silly and don't specify an extension in the glob pattern
+		// filter out unwanted file extensions
+		// for silly users that don't specify an extension in the glob pattern
 		paths = paths.filter(x => {
-			const ext = path.extname(x);
-			return ext === '.js' || ext === '.jsx';
+			// Remove dot before the actual extension
+			const ext = path.extname(x).replace('.', '');
+			return opts.extensions.indexOf(ext) > -1;
```
sindresorhus: Use Array#includes

readme.md
```diff
@@ -197,6 +199,12 @@ Type: `Array`, `string`
 
 Use one or more [shareable configs](http://eslint.org/docs/developer-guide/shareable-configs.html) to override any of the default rules (like `rules` above).
 
+### extensions
+
+Type: `Array`, `string`
```
sindresorhus: Should only be Array

sindresorhus:
> Should I change the default pattern to **/*? 

👍

sindresorhus:
And PR title needs to be updated

index.js
sindresorhus: Agh, right. Forgot that.

Use `!== -1`

sindresorhus:
Looks great! Thanks @addobandre :)

PR #157, opened 2016-10-23
[WIP] Add watch mode

### LINKED ISSUE #136: Watch mode
Didn't find an option for that, and would be awesome.

vadimdemedes:
Fixes #136.

Todo:
- [ ] Handle case when chokidar native dependency wasn't installed
- [ ] Add tests
- [ ] Update readme

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -19,6 +19,8 @@ const path = require('path');
 const spawn = require('child_process').spawn;
 const updateNotifier = require('update-notifier');
 const getStdin = require('get-stdin');
+const chokidar = require('chokidar');
+const chalk = require('chalk');
 const meow = require('meow');
 const formatterPretty = require('eslint-formatter-pretty');
 const xo = require('./');
@@ -43,6 +45,7 @@ const cli = meow(`
 	  --open          Open files with issues in your editor
 	  --quiet         Show only errors and no warnings
 	  --extension     Additional extension to lint [Can be set multiple times]
+	  --watch         Re-run XO when files change
 
 	Examples
 	  $ xo
@@ -65,7 +68,8 @@ const cli = meow(`
 		'compact',
 		'stdin',
 		'fix',
-		'open'
+		'open',
+		'watch'
 	]
 });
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
How would this interact with other watch modes, like the AVA watch mode? What if I want to both lint and run tests with watching at the same time? Why would you use watch mode over a editor plugin which runs live and shows error inline?

vdemedes:
I see myself using `xo --watch` before submitting a PR, when I finished writing the code and now need to fix all the lint errors in all modified files. I constantly lose time during this routine. Fix some errors, run xo, fix some more, run xo, repeat while true.

> Why would you use watch mode over a editor plugin which runs live and shows error inline?

As for me, I prefer as little editor plugins and as minimal UI as possible. This is just my preference though.

sindresorhus:
You want a watch mode to make it easier to fix lint violations, but an editor plugin makes it even easier. Instead of having to go back and forth between the terminal and editor, check remaining errors, find the corresponding file and line and fix it, you could just have that info right inside your editor. With the Atom plugin, you can customize exactly how much noise you want. Don't want inline highlighting, go for just gutter marks:

screen shot 2016-11-05 at 21 28 11

vdemedes:
Atom plugin is surely awesome, but I use Sublime and I always prefer CLI over editor plugins.

sindresorhus:
@vadimdemedes Have you tried the Sublime plugin? It's pretty fast and nice to use. Give it a try for a week and see if you still feel the same.

Anyways. Could you make a separate CLI for this instead? For example `xo-watch`. Also, there must be good generic watchers out there that you could use too. All you're doing here is calling XO with file paths, so you should be able to do that using a generic watcher CLI tool too.

vadimdemedes:
@sindresorhus Nah, I love the CLI.

> Could you make a separate CLI for this instead? For example xo-watch. All you're doing here is calling XO with file paths, so you should be able to do that using a generic watcher CLI tool too.

Sure, sounds reasonable.

zzyyxxww:
@sindresorhus I know that this is a 3 year old comment, but I'll do my take on it.
> How would this interact with other watch modes, like the AVA watch mode? What if I want to both lint and run tests with watching at the same time? 

Linting and testing are two different things, so I like running them on 2 different shells, which the popular software tmux accomodates for.   

> Why would you use watch mode over a editor plugin which runs live and shows error inline?

I'm a vim user so a linter running async on the editor is not applicable to it, and I prefer CLIs also.

My problem is that the watcher I use for linting is `eslint-watch`, which is suboptimal because it spawns a new node process everytime.

I think it makes total sense to have a `watch` mode for xo. There is one on ava, I think it'd be nice to have one here also. I could provide a PR.

PR #158, opened 2016-10-30
Ignore directories when accepting empty extensions (fixes #156)

### LINKED ISSUE #156: Ignored paths are not respected when linting files without extension
I am trying to upgrade a project to XO 0.17 (https://github.com/zeit/now/pull/79), and in version 0.16, it works fine. When bumping to 0.17, some folders that were previously ignored are now being linted.

I have overridden the `extensions` option to `["", ".js"]` in order to include (bin) files without extensions. At the same test, I have a ignored fixtures folder at `test/_fixtures/`, while I am linting `test/`.

What happens is that, at [this point](https://github.com/sindresorhus/xo/blob/master/index.js#L37-L41), the ignores is properly set and you get all the good files and directories. But, because I have overridden the extensions to allow `""`, directories are not filtered out. Then when those paths get passed to ESLint, it gets run on the `test/` folder, which causes all of its subdirectories, including `test/_fixtures` to be linted too.

I think that adding the `no-dir` option to `globby` would solve the problem, but I am pretty tired so I don't know if that solves it properly without creating new problems :sweat_smile:

jfmengels:
Ignore directories when accepting empty extensions (fixes #156)

Let me know if I need to change something :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ exports.lintFiles = (patterns, opts) => {
 		patterns = '**/*';
 	}
 
-	return globby(patterns, {'ignore': opts.ignores, 'no-dir': true}).then(paths => {
+	return globby(patterns, {ignore: opts.ignores, nodir: true}).then(paths => {
 		// filter out unwanted file extensions
 		// for silly users that don't specify an extension in the glob pattern
 		paths = paths.filter(x => {
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ exports.lintFiles = (patterns, opts) => {
 		patterns = '**/*';
 	}
 
-	return globby(patterns, {ignore: opts.ignores}).then(paths => {
+	return globby(patterns, {'ignore': opts.ignores, 'no-dir': true}).then(paths => {
 		// filter out unwanted file extensions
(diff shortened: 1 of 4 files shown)


jfmengels:
Fixed the option name, but I can't seem to re-create a failing test case, so I just removed them. Let me know if you really want tests, if so I'll try to get back to it later. This works fine for the repo that caused problems.

sindresorhus:
I would prefer a test so to make sure there's no future regression. There's a lot of moving parts in XO and easy to introduce silent bugs.

sindresorhus:
ping @jfmengels :)

ntwcklng:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

marionebl:
I pushed the changes of this PR on a freshly rebased branch and added a relevant test case: https://github.com/sindresorhus/xo/tree/jfmengels/no-dir

Note I moved the case to `test/api.js`. The case fails with `nodir: false`.

marionebl:
@jfmengels @sindresorhus How do we want to go forward on this?

sindresorhus:
@marionebl Can you do a PR with your branch and include the change from this PR?

PR #162, opened 2016-11-12
Support plugin config in the `extends` option

denniss17:
## Current issue
XO allows to specify sharable configs to override any of the default rules [link](https://github.com/sindresorhus/xo#extends). However, in ESLint it is allowed to use configs shared by plugins (prefixed with `eslint-plugin-`, next to shared configs (prefixed with `eslint-config-` and absolute/relative paths. These plugin configs are prefixed with `plugin:`, [see here](http://eslint.org/docs/user-guide/configuring#using-the-configuration-from-a-plugin).

## Contribution
This PR allows to specify such plugin configs in the XO configuration, by checking whether the name starts with `plugin:` and only prepending `eslint-config-` if that is not the case.

### DIFF
--- a/options-manager.js
+++ b/options-manager.js
@@ -145,6 +145,11 @@ function buildConfig(opts) {
 				return name;
 			}
 
+			// don't do anything if it's a config from a plugin
+			if (name.indexOf('plugin:') === 0) {
+				return name;
+			}
+
 			if (!name.includes('eslint-config-')) {
 				name = `eslint-config-${name}`;
 			}
--- a/options-manager.js
+++ b/options-manager.js
@@ -146,7 +146,7 @@ function buildConfig(opts) {
 			}
 
 			// don't do anything if it's a config from a plugin
-			if (name.indexOf('plugin:') === 0) {
+			if (name.startsWith('plugin:')) {
 				return name;
 			}
 
--- a/test/options-manager.js
+++ b/test/options-manager.js
@@ -88,6 +88,14 @@ test('buildConfig: settings', t => {
 	t.deepEqual(config.baseConfig.settings, settings);
 });
 
+test('buildConfig: extends', t => {
+	const extendsOption = ['plugin:foo/bar', 'eslint-config-foo-bar', 'foo-bar-two'];
+	const config = manager.buildConfig({extends: extendsOption});
+	t.is(config.baseConfig.extends[config.baseConfig.extends.length - 3], 'plugin:foo/bar');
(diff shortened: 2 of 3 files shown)


options-manager.js
```diff
@@ -145,6 +145,11 @@ function buildConfig(opts) {
 				return name;
 			}
 
+			// don't do anything if it's a config from a plugin
+			if (name.indexOf('plugin:') === 0) {
```
sindresorhus: Use `String#startsWith`

test/options-manager.js
```diff
@@ -88,6 +88,14 @@ test('buildConfig: settings', t => {
 	t.deepEqual(config.baseConfig.settings, settings);
 });
 
+test('buildConfig: extends', t => {
+	const extendsOption = ['plugin:foo/bar', 'eslint-config-foo-bar', 'foo-bar-two'];
+	const config = manager.buildConfig({extends: extendsOption});
+	t.is(config.baseConfig.extends[config.baseConfig.extends.length - 3], 'plugin:foo/bar');
+	t.is(config.baseConfig.extends[config.baseConfig.extends.length - 2], 'cwd/eslint-config-foo-bar');
+	t.is(config.baseConfig.extends[config.baseConfig.extends.length - 1], 'cwd/eslint-config-foo-bar-two');
```
sindresorhus: Too much boilerplate. Put `config.baseConfig.extends` in a variable and use it.

test/options-manager.js
```diff
@@ -88,6 +88,15 @@ test('buildConfig: settings', t => {
 	t.deepEqual(config.baseConfig.settings, settings);
 });
 
+test('buildConfig: extends', t => {
+	const extendsOption = ['plugin:foo/bar', 'eslint-config-foo-bar', 'foo-bar-two'];
+	const config = manager.buildConfig({extends: extendsOption});
+	const extendsConfig = config.baseConfig.extends;
+	t.is(extendsConfig[extendsConfig.length - 3], 'plugin:foo/bar');
+	t.is(extendsConfig[extendsConfig.length - 2], 'cwd/eslint-config-foo-bar');
+	t.is(extendsConfig[extendsConfig.length - 1], 'cwd/eslint-config-foo-bar-two');
```
sindresorhus: Actually, would be even nicer to just `Array#slice()` off the items we want an do an `t.deepEqual`.

sindresorhus:
Thanks :)

denniss17:
No problem :+1:

Issue #163, opened 2016-11-15
forbidding --fix with --stdin makes it incompatible with vim-autoformat


gautaz:
[vim-autoformat](https://github.com/Chiel92/vim-autoformat) documentation [states](https://github.com/Chiel92/vim-autoformat#how-can-i-change-the-behaviour-of-formatters-or-add-one-myself):
>The formatprogram must read the unformatted code from the standard input, and write the formatted code to the standard output.

I guess this limitation (forbidden `--fix --stdin`) comes from `eslint` but I still ask in case there is anything that can be done to work around the limitation on `xo`'s side.

sindresorhus:
I just followed [ESLint's lead](https://github.com/eslint/eslint/issues/5393), but in hindsight it doesn't make sense. Added in https://github.com/sindresorhus/xo/pull/164. Please try it out and report back :)

gautaz:
@sindresorhus Just tried your `stdin-fix` branch, works like a charm with `vim-autoformat`!
Thanks! :-)

sindresorhus:
@gautaz Cool. Would you be willing to submit a short instruction on how to use the VIM autoformat and XO? Here: https://github.com/sindresorhus/xo#vim

gautaz:
Done in #165, hope it is enough.
Additional configuration might be needed in case you want to use the XO's module from the current project but it seems a bit out of scope (for an example see this [autoformat config](https://github.com/gautaz/dotfiles/blob/master/neovim/.config/nvim/modules/autoformat.vim) using these [utilities](https://github.com/gautaz/dotfiles/blob/master/neovim/.config/nvim/modules/utilities.vim)).

sindresorhus:
> Additional configuration might be needed in case you want to use the XO's module from the current project but it seems a bit out of scope

XO already handles that. If you call the global XO binary and there's one in your project, it will defer to that one instead.

gautaz:
> XO already handles that. If you call the global XO binary and there's one in your project, it will defer to that one instead.

Gosh! If only all linters could do that ;-).

PR #165, opened 2016-11-16
document xo's usage from vim-autoformat

gautaz:
Instructions to use XO within [vim-autoformat](https://github.com/Chiel92/vim-autoformat) (related to #163).


sindresorhus:
Thank you :)

PR #169, opened 2016-11-26
feat: apply `.ignores` to `lintText`

### LINKED ISSUE #168: lintText does not apply ignores config
`lintText` does not apply the [ignores](https://github.com/sindresorhus/xo#ignores) config as opposed to `lintFiles`. 

Given `lintText` has `opts.filename` and `opts.cwd` available we could compute if a given file should be ignored. 

Don't know if this is less intuitive, though. What do you think @sindresorhus?

---

* https://github.com/sindresorhus/atom-linter-xo/issues/22

marionebl:
- add multimatch-powered short-circuit to `lintText`
- add supporting test cases
- throw for opts.ignores without opts.filename
-  use default ignores when opts.filename is given
- closes #168

BREAKING CHANGES:
* lintText now uses the same default ignores as lintFiles with opts.filename
  * **/node_modules/**
  * **/bower_components/**
  * coverage/**
  * {tmp,temp}/**
  * **/*.min.js
  * **/bundle.js
  * fixtures{-*,}.{js,jsx}
  * fixture{s,}/**
  * {test,tests,spec,__tests__}/fixture{s,}/**
  * vendor/**
  * dist/**
* lintText now throws when providing opts.ignores without opts.filename

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,8 @@
 'use strict';
 const path = require('path');
-const arrayEqual = require('array-equal');
 const eslint = require('eslint');
 const globby = require('globby');
+const isEqual = require('lodash.isequal');
 const multimatch = require('multimatch');
 const optionsManager = require('./options-manager');
 
@@ -22,8 +22,8 @@ exports.lintText = (str, opts) => {
 	opts = optionsManager.buildConfig(opts);
 	const defaultIgnores = optionsManager.getIgnores({}).ignores;
 
-	if (opts.ignores && !arrayEqual(defaultIgnores, opts.ignores) && typeof opts.filename !== 'string') {
-		throw new Error(`xo: opts.filename must be string when providing opts.ignores. Received value "${opts.filename}" of type "${typeof opts.filename}" for opts.filename.`);
+	if (opts.ignores && !isEqual(defaultIgnores, opts.ignores) && typeof opts.filename !== 'string') {
+		throw new Error('The `ignores` option requires the `filename` option to be defined.');
 	}
 
 	if (opts.ignores && opts.ignores.length > 0 && opts.filename) {
--- a/index.js
+++ b/index.js
@@ -2,16 +2,35 @@
 const path = require('path');
 const eslint = require('eslint');
 const globby = require('globby');
+const multimatch = require('multimatch');
 const optionsManager = require('./options-manager');
 
 exports.lintText = (str, opts) => {
 	opts = optionsManager.preprocess(opts);
+
+	if (opts.cwd && opts.filename) {
+		const filename = path.relative(opts.cwd, opts.filename);
+
+		if (multimatch([filename], opts.ignores)) {
+			return {
+				errorCount: 0,
+				warningCount: 0,
+				results: [{
+					errorCount: 0,
+					filePath: filename,
+					messages: [],
+					warningCount: 0
+				}]
+			};
+		}
+	}
 
 	if (opts.overrides && opts.overrides.length > 0) {
 		const overrides = opts.overrides;
 		delete opts.overrides;
 
 		const filename = path.relative(opts.cwd, opts.filename);
+
 		const foundOverrides = optionsManager.findApplicableOverrides(filename, overrides);
 		opts = optionsManager.mergeApplicableOverrides(opts, foundOverrides.applicable);
 	}
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const path = require('path');
+const arrayEqual = require('array-equal');
 const eslint = require('eslint');
 const globby = require('globby');
 const multimatch = require('multimatch');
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -2,16 +2,35 @@
...
+				warningCount: 0,
+				results: [{
+					errorCount: 0,
+					filePath: filename,
+					messages: [],
+					warningCount: 0
+				}]
+			};
+		}
+	}
```
sindresorhus: This needs to take into account the overrides logic and should be after the below overrides logic. Right now it wouldn't work if you used config overrides that enabled it again. Should also have a test for that to ensure it works.

index.js
```diff
@@ -2,16 +2,35 @@
 const path = require('path');
 const eslint = require('eslint');
 const globby = require('globby');
+const multimatch = require('multimatch');
 const optionsManager = require('./options-manager');
 
 exports.lintText = (str, opts) => {
 	opts = optionsManager.preprocess(opts);
 
+	if (opts.cwd && opts.filename) {
```
sindresorhus: To be user-friendly, I think we should throw an error if the user of `lintText` tries to specify the `ignore` option, but didn't supply `cwd` and `filename`.

marionebl: Done! Guess this is a breaking change as passing `ignores` before these changes did nothing.

index.js
marionebl: Done. Amounts to another breaking change as the default ignores now take effect

marionebl: Test: https://github.com/sindresorhus/xo/pull/169/files#diff-9265a69d5e81bfaab4edbd417a33a626R49

test/api.js
```diff
@@ -20,6 +20,55 @@ test('.lintText() - `esnext` option', t => {
...
+test('.lintText() - `ignores` option without cwd', t => {
+	const result = fn.lintText(`'use strict'\nconsole.log('unicorn');\n`, {
+		filename: 'ignored/index.js',
+		ignores: ['ignored/**/*.js']
+	});
+	t.is(result.errorCount, 0);
+	t.is(result.warningCount, 0);
+});
+
+test('.lintText() - respect overrides', t => {
```
marionebl: As requested: https://github.com/sindresorhus/xo/pull/169/files#r89684310

index.js
```diff
@@ -12,11 +14,34 @@ exports.lintText = (str, opts) => {
...
 		const filename = path.relative(opts.cwd, opts.filename);
+
 		const foundOverrides = optionsManager.findApplicableOverrides(filename, overrides);
 		opts = optionsManager.mergeApplicableOverrides(opts, foundOverrides.applicable);
 	}
 
 	opts = optionsManager.buildConfig(opts);
+	const defaultIgnores = optionsManager.getIgnores({}).ignores;
+
+	if (opts.ignores && !arrayEqual(defaultIgnores, opts.ignores) && typeof opts.filename !== 'string') {
```
sindresorhus: I don't really understand the purpose of the `!arrayEqual(defaultIgnores, opts.ignores)` check. ?

I would also prefer using `lodash.isequal` instead as it's more general and could be used for other deep checks.

It should also check that the `cwd` option is defined.

index.js
```diff
@@ -12,11 +14,34 @@ exports.lintText = (str, opts) => {
...
+		throw new Error(`xo: opts.filename must be string when providing opts.ignores. Received value "${opts.filename}" of type "${typeof opts.filename}" for opts.filename.`);
```
sindresorhus: This is a bit too verbose. People will most likely never hit this as we're the only ones using the API. I also don't see the point of the `xo:` prefix. It's clear from the stack trace where the error is coming from.

I'd prefer something like:

```
> The `ignores` option requires the `filename` and `cwd` option to be defined.
```

index.js
marionebl: - [x] `array-equal` => `lodash.isequal`
- the check is there to test if the user provided ignores via `.ignores` or `.overrides[index].ignores`. `opts.ignores` is always defined

test/api.js
```diff
@@ -20,6 +20,55 @@ test('.lintText() - `esnext` option', t => {
...
+	const result = fn.lintText(`'use strict'\nconsole.log('unicorn');\n`, {
+		filename: 'ignored/index.js',
+		ignores: ['ignored/**/*.js'],
+		overrides: [
+			{
+				files: ['ignored/**/*.js'],
+				ignores: []
```
marionebl: Done

marionebl:
@sindresorhus Could you have another look?

sindresorhus:
> the check is there to test if the user provided ignores via .ignores or .overrides[index].ignores. opts.ignores is always defined

Feels a bit overkill. How about just putting a variable at the top with:

```js
const hasIgnores = Array.isArray(opts.ignores);
```

And using that further down.

marionebl:
That would not handle ignores in `opts.overrides` then

sindresorhus:
Right, forgot about the `ignores` in overrides. I'm still not getting how `!isEqual(defaultIgnores, opts.ignores)` works though (might be because it's late here and I'm half asleep). Wouldn't it have a false-positive if both `defaultIgnores` were `opts.ignores` empty arrays?

marionebl:
Not sure if the current approach is good enough if it is hard to grasp. Let's defer this until tomorrow and have some sleep 💤

marionebl:
Coming back to this: Given there was a case with two empty array this would do the right thing and skip the error-producing code, right?

marionebl:
@sindresorhus Can we move forward?

sindresorhus:
Yes, sorry about the delay. Looks great. Thank you :)

Issue #182, opened 2017-02-12
Making default extenders optional


camilleriluke:
My team and I have our own curated list of eslint configs that we use different extenders based on file patterns in our codebase. One thing that I found a bit not so configurable is the hardcoded extenders that are shipped with the plugin ([source](https://github.com/sindresorhus/xo/blob/master/options-manager.js#L37)).

I understand that they could be turned off one by one but that is quite a maintenance cost, if there is another way how we could achieve this please let me know. 

## Proposed update
Alternative I would like to put forward a suggestion that could solve this. 

Provide a flag (`skip-default-extenders`) that when turned on it will skip the default extenders. This flag will be `false` by default, so it will still respect the zero-config mantra of the plugin

sindresorhus:
Why even use XO if you want to skip all the built-in rules? Seems like just using ESLint would be better, or am I missing something?

camilleriluke:
I'm using it mostly for the glob based linting.

marionebl:
I feel this should not be part of xo. If there is a use case for the globbing/ignoring/overrides in xo without its rulesets the functionality could be moved to a separate module and cli tool.

sindresorhus:
I've been thinking about this for a while, and while it's a small change, it changes the scope of XO dramatically and opens the door to lots of other requests.

It also seems like ESLint will soon have support for glob-based config anyhow: https://github.com/eslint/eslint/pull/8081

camilleriluke:
Thanks for your time anyway 👍

PR #183, opened 2017-02-12
Making default extenders optional

camilleriluke:
Related issue: #182

## Changes
Provide a flag (`skip-default-extenders`) that when turned on it will skip the default extenders. This flag will be `false` by default, so it will still respect the zero-config mantra of the plugin


sindresorhus:
Sorry. Going to pass on this. https://github.com/sindresorhus/xo/issues/182#issuecomment-284515901

PR #184, opened 2017-02-16
Add yarn-cache to ignored paths

hugomrdias:
for some CIs we need to change yarn-cache to the working directory so it can be cache between builds

### DIFF
--- a/options-manager.js
+++ b/options-manager.js
@@ -22,7 +22,7 @@ const DEFAULT_IGNORE = [
 	'{test,tests,spec,__tests__}/fixture{s,}/**',
 	'vendor/**',
 	'dist/**',
-	'yarn-cache/**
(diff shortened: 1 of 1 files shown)


marionebl:
The default yarn cache does not reside in `process.cwd()`, correct? 
`xo` has to be conservative with additional default ignores to avoid surprises for users, so I'd vote against adding this special case.

You can ignore the directory via the [ignore flag](https://github.com/sindresorhus/xo#usage) instead:

```
xo --ignore yarn-cache/**/*
```

or the [ignores option](https://github.com/sindresorhus/xo#ignores):

```
{
  "xo": {
    "ignores": ["yarn-cache/**/*"]
   }
}
```

Alternatively you can add it to `.gitignore` as `xo` evaluates it, too.

PR #192, opened 2017-03-15
Cache node_modules to speed up builds on Travis

j-f1:
Should save about 30 seconds per build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: node_js
 node_js:
   - '6'
   - '4'
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


sindresorhus:
I'm not a big fan of caching the build. Yes, it saves some time, but it also means that XO could be broken by a sub-dependency and I wouldn't know. I prefer having a fresh dependency-tree every time I run Travis. Correctness > Speed.

For example: https://github.com/zeit/hyper/commit/67cefc50a8c2e164ec90bb4d36059c58f353bb8c

PR #194, opened 2017-03-17
Ignore directories when receiving an empty extension

marionebl:
* Builds up on work of @jfmengels 
* Add relevant API tests

---

Will squash after review


sindresorhus:
Thanks for picking up this :)

PR #195, opened 2017-03-17
Handle negative gitignore patterns correctly

### LINKED ISSUE #154: Negative .gitignore rules cause xo to pass everything
There is a regression in version 0.17, which in the worst case leads to `xo` ignoring all files and returning exit code 0, even though there are issues in the code. For instance, `xo@0.17` will pass all tests in a `react-native` project no matter what because the `.gitignore` file has entries that start with `!`.
## Test case

Here is a simple example project to demonstrate the issue:

<hr>

`package.json`

``` json
{
  "name": "xo-bug-demo",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "xo"
  },
  "author": "Jonny <jonathanburger11@gmail.com> (http://jonny.io)",
  "license": "ISC",
  "devDependencies": {
    "xo": "^0.17.0"
  }
}
```

<hr>

`index.js`

``` js
console.log('No semicolon – xo should fail')
```

<hr>

`.gitignore`

```
some_folder
!some_folder/but_not_this_file
```

<hr>
## Expected result

```
❯ xo

  index.js:1:45
  ✖  1:45  Missing semicolon.  semi

  1 error
```
## Actual result

```
❯ xo

```

xo returns no error even though the semicolon is missing.
Running with `xo@0.16` will yield the correct result.
## Cause

Starting in `0.17`, files in `.gitignore` are also ignored in `xo`. But there is a difference in how `git` and `globby` parse the ignore list.  
In `.gitignore`, `!a/b.js` means: "Do not ignore this file".  
In `globby` and `glob`, passing `!a/b.js` to the ignore option means: "Do ignore all files except this one"

marionebl:
*  split out getGitIgnores from getIgnores
*  adapt test cases accordingly
*  extend gitignore test case with negative pattern
*  invert gitignore patterns and feed sotred as glob instead of opts.ignore

fix #154

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,10 +26,12 @@ exports.lintText = (str, opts) => {
 		throw new Error('The `ignores` option requires the `filename` option to be defined.');
 	}
 
-	if (opts.ignores && opts.ignores.length > 0 && opts.filename) {
+	if (opts.filename) {
+		const gitIgnores = optionsManager.getGitIgnores(opts);
 		const filename = path.relative(opts.cwd, opts.filename);
+		const glob = [filename].concat(gitIgnores);
 
-		if (multimatch([filename], opts.ignores).length > 0) {
+		if (multimatch(glob, opts.ignores).length > 0) {
 			return {
 				errorCount: 0,
 				warningCount: 0,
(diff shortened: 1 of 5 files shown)


sindresorhus:
Yay! Thanks for fixing this. It's been eating at me for a long time.

sindresorhus:
@schnittstabil @JonnyBurger I already merged this as it looks good to me, but your review would still be welcome :)

schnittstabil:
@marionebl Thanks for exploring this.

Hmm, as far as I can see, negative `.gitignore` entries will always be linted from now on:

```
# .gitignore
!bar.js
```

```js
xo.lintFiles('foo.js'); // should not lint 'bar.js', should it?
```

JonnyBurger:
@schnittstabil @marionebl You are right, I was able to reproduce the bug you described. 
I added a failing test to demonstrate: #199.

PR #199, opened 2017-03-19
Add failing test for negative .gitignore rules

JonnyBurger:
This test demonstrates the bug that negative entries in `.gitignore` now get linted even if we don't include them in the list of files we want linted.

See: https://github.com/sindresorhus/xo/pull/195#issuecomment-287563834


JonnyBurger:
Stumbled upon another issue, a test on Windows is not actually working:

```js
test('ignore files in .gitignore', async t => {
	const cwd = path.join(__dirname, 'fixtures/gitignore');

	try {
		const result = await execa('../../../cli.js', ['--no-local'], {cwd});
	} catch (err) {
		t.is(err.stdout.indexOf('foo.js'), -1);
		t.true(err.stdout.indexOf('bar.js') !== -1);
	}
        // No error gets thrown, therefore the test passes even though it shouldn't.
});
```

That is why my 'failing' test ended up passing and that is why the build is broken.  
Any help from more knowledgable people is appreciated.

test/cli.js
```diff
@@ -52,13 +52,26 @@ test('ignore files in .gitignore', async t => {
...
 	try {
-		await execa('../../../cli.js', ['--no-local'], {cwd});
+		const result = await execa('../../../cli.js', ['--no-local'], {cwd});
+		console.log(result, result.stdout, result.stderr);
+		t.fail();
 	} catch (err) {
 		t.is(err.stdout.indexOf('foo.js'), -1);
 		t.true(err.stdout.indexOf('bar.js') !== -1);
 	}
 });
```
schnittstabil:
```diff
-       try {
-               await execa('../../../cli.js', ['--no-local'], {cwd});
-       } catch (err) {
-               t.false(err.stdout.includes('foo.js'));
-               t.true(err.stdout.includes('bar.js'));
-       }
+       const err = await t.throws(execa('../../../cli.js', ['--no-local'], {cwd}));
+       t.false(err.stdout.includes('foo.js'));
+       t.true(err.stdout.includes('bar.js'));
```

test/cli.js
```diff
@@ -52,13 +52,26 @@ test('ignore files in .gitignore', async t => {
...
+test.failing('negative gitignores', async t => {
+	const cwd = path.join(__dirname, 'fixtures/negative-gitignore');
+
+	try {
+		await execa(`../../../cli.js`, [`${cwd}/bar.js`, '--no-local'], {cwd});
+		t.fail();
+	} catch (err) {
+		t.is(err.stdout.indexOf('foo.js'), -1, 'Should not lint foo.js');
+	}
+});
```
schnittstabil:
```js
const err = await t.throws(execa(`../../../cli.js`, [`${cwd}/bar.js`, '--no-local'], {cwd});
t.is(err.stdout.indexOf('foo.js'), -1, 'Should not lint foo.js');
```

JonnyBurger: This is cool! I didn't know that you could `await t.throws` and then make further assertions with the error. Thanks and I changed it!

sindresorhus:
Fixed by https://github.com/sindresorhus/xo/pull/204.

sindresorhus:
Thanks for the tests @JonnyBurger 🙌

Issue #201, opened 2017-03-26
Recursive glob patterns


CrashyBang:
Running `0.18.1` on Ubuntu 16.10 with Node `v7.7.1`.

I have the following command: `xo resources/**/*.js`.

I am trying to use it on the following structure:

```
├── admin
│   ├── assets
│       ├── main.js
└── webpack
    ├── config.js
```

Now that reports all linting failues in `config.js` but not `main.js` if I change that command to: `xo resources/**/**/*.js`. it now works but obviously I don't want to have to add a `/**` everytime I go one directory deeper.

I have looked at [globby#24](https://github.com/sindresorhus/globby/pull/24) and [cpy#10](https://github.com/sindresorhus/cpy/issues/10) but neither seemed to really address the issue.

Is this by intentition or is this a bug?

In an ideal world `xo resources/**/*.js` would traverse all folders underneath `resources` recursively.

Cheers.

Qix-:
What happens when you use `xo 'resources/**/*.js'` (in single quotes)?

CrashyBang:
Hey @Qix- that seems to have fixed it.

I am running as an npm script: `"lint:scripts": "xo 'resources/**/*.js'",`.

Cheers.

Qix-:
Yep - without the single quotes, it's your shell trying to expand the double star, not XO. To get it to work otherwise, you'd need `shopt -s globstar` in your `.bash_profile`. But the single quotes is a better solution since it works for everyone. :)

PR #204, opened 2017-04-02
Do not always lint matches on negative gitignore patterns

marionebl:
* fixes fallout of #195
* builds on #201, #202
* removes parse-gitignore in favor of node-ignore
* thanks @JonnyBurger for providing the first failing test case

Please review in detail, this is my first fast-and-furious attempt on fixing the issues around gitignore parsing

--- 
~~There appears to be some win32 related issue in `getGitIgnoreFilter` still.~~


index.js
```diff
@@ -56,17 +56,19 @@ exports.lintFiles = (patterns, opts) => {
...
 		// Filter out unwanted file extensions
 		// for silly users that don't specify an extension in the glob pattern
-		paths = paths.filter(x => {
-			// Remove dot before the actual extension
-			const ext = path.extname(x).replace('.', '');
-			return opts.extensions.indexOf(ext) !== -1;
-		});
+
+		paths = paths
+			.filter(ignoreFilter)
```
schnittstabil: Is there a good reason to filter out `.gitignore` patterns _after_ `globby`?

I believe this would be a major performance hit, e.g. if we conisder the `.gitignore` file below, then all files in `node_modules` will be globbed but filtered afterwards:

```
# .gitignore
node_modules
```

marionebl: I think this is the easiest way to avoid the bug introduced with #195 where every negative gitignore pattern is **always** linted, e.g.

```
❯ tree -a negative-gitignore/
negative-gitignore/
├── .gitignore
├── bar.js
└── foo.js
```

```
❯ cat negative-gitignore/.gitignore 
*.js
!foo.js
```

```
xo negative-gitignore/bar.js # unexpectedly lints foo.js
```

marionebl:
> > * removes parse-gitignore in favor of node-ignore

> Mhm, why?

Did this because your fixes where unmerged yesterday. I am not attached to that change, though - will reintroduce `parse-gitignore` to make this change way smaller

schnittstabil:
> Edit: `node-ignore` does give us all the rather involved interpretation logic attached to making negative gitignore patterns work correctly.

Indeed, `node-ignore` seems more appropriate.

Btw, have you considered this use-case as well:
```bash
$ tree -a
.
├── dir
│   ├── .gitignore
│   └── unicorn.js
└── .gitignore

1 directory, 3 files

$ cat .gitignore 
dir

$ cat dir/.gitignore 
!unicorn.js

$ git init && git status -a
?? .gitignore
```

index.js
sindresorhus: I would prefer to filter out `.gitignore` before `globby` for performance reasons, as @schnittstabil mention, but I'd be ok with landing this now to fix the bugs and open an issue about improving the performance later.

options-manager.js
```diff
@@ -225,25 +228,35 @@ const getIgnores = opts => {
...
+	globby.sync('**/.gitignore', {
+		ignore: ignores,
+		cwd: cwd
 	})
-	.reduce((a, b) => a.concat(b), []);
+	.map(file => {
+		const fileName = path.join(cwd, file);
+		const base = slash(path.relative(cwd, path.dirname(fileName)));
+
+		const lines = fs.readFileSync(fileName)
```
marionebl: Done!

marionebl:
> > Edit: node-ignore does give us all the rather involved interpretation logic attached to making negative gitignore patterns work correctly.
Indeed, node-ignore seems more appropriate.
> 
> Btw, have you considered this use-case as well:
> ```bash
> $ tree -a
> .
>  ├── dir
> │   ├── .gitignore
> │   └── unicorn.js
> └── .gitignore
> 
> 1 directory, 3 files
> 
> $ cat .gitignore 
> dir
> 
> $ cat dir/.gitignore 
> !unicorn.js
> 
> $ git init && git status -a
> ?? .gitignore
> ```

Yep, that works. Hard to add a test case though.

sindresorhus:
Great work @marionebl. And thanks for reviewing @schnittstabil. Glad we could finally get the gitignore stuff right.

sindresorhus:
@marionebl Can you open an issue about https://github.com/sindresorhus/xo/pull/204/files#r111681661 ?

PR #211, opened 2017-04-24
Adds yarn.lock to gitignore, adds test for local xo and xo --init

OmgImAlexis:
As title suggests, if there's a better way to add these test please let me know. I noticed the coverage was less than 100% so I'm trying to add more tests to help fix that. 👍


OmgImAlexis:
The only test that seems to be failing isn't from anything I've changed.
"ignore files in .gitignore"

OmgImAlexis:
I'm guessing to hit LN12 of cli.js you need to be using a global version of xo?

test/cli.js
```diff
@@ -69,3 +69,12 @@ test('quiet option', async t => {
 	const err = await t.throws(execa('../cli.js', ['--no-local', '--quiet', '--reporter=compact', filepath]));
 	t.is(err.stdout.indexOf('warning'), -1);
 });
+
+test('init option', async t => {
+	const filepath = await tempWrite('{}', 'package.json');
+	await execa(path.join(__dirname, '../cli.js'), ['--init'], {
+		cwd: filepath.replace('package.json', '')
```
sindresorhus:
```js
cwd: path.dirname(filepath)
```

test/cli.js
```diff
@@ -69,3 +69,12 @@ test('quiet option', async t => {
...
+	});
+	const packagejson = await fs.readFileSync(filepath, 'utf8');
```
sindresorhus: You're await'ing a synchronous function.

test/cli.js
```diff
@@ -69,3 +69,12 @@ test('quiet option', async t => {
...
+	t.regex(JSON.stringify(JSON.parse(packagejson)), /{"scripts":{"test":"xo"},"devDependencies":{"xo":"\^[0-9]+\.[0-9]+\.[0-9]+"}}/);
```
sindresorhus: Do it the other way around. Parse the JSON and do a `deepEqual` assertion on it.

sindresorhus:
> I'm guessing to hit LN12 of cli.js you need to be using a global version of xo?

Yes, or somehow imitate it. I don't think it's worth the trouble though.

test/cli.js
OmgImAlexis: Wouldn't that mean only scripts could be checked since we can't hardcode the version number or should I grab the version number from `../package.json` and test against that?

sindresorhus: > should I grab the version number from ../package.json and test against that?

👍

schnittstabil:
> I'm guessing to hit LN12 of cli.js you need to be using a global version of xo?

@OmgImAlexis #218 adds some very simple tests for that (using `--no-local`) :grin:

PR #221, opened 2017-05-20
Open the errors first then warnings

### LINKED ISSUE #219: `--open` should open the first error, not warning
I have warnings in many projects and it's annoying that it opens at the warning instead of the error. If there are no errors, it should go to the first warning. I think we could just sort the errors first.

Should not be done until https://github.com/sindresorhus/xo/pull/218 is merged.

tdeschryver:
Fixes #219


main.js
```diff
@@ -77,20 +77,20 @@ const log = report => {
 	process.exit(report.errorCount === 0 ? 0 : 1);
 };
 
+const files = (report, predicate) => report.results
+	.filter(predicate)
```
schnittstabil: `sort` seems more appropriate to me, i.e.:

0. assert `warningCount` and `errorCount` are not `undefined` [*]
1. filter out all results without errors or warnings
2. sort: At first, errors before warnings. Secondly,  _(error|warning)Count_ in descending order.

[*] I do not know, if this is actually needed, maybe `warningCount` and `errorCount` are always numbers.

For example:
```js
.map(result => Object.assign({warningCount: 0, errorCount: 0}, result))
.filter(result => result.errorCount + result.warningCount > 0)
.sort((a, b) => a.errorCount + b.errorCount > 0 ? (b.errorCount - a.errorCount) : (b.warningCount - a.warningCount))
```

tdeschryver: 0: seems fair to me, better safe than sorry
1-2: I had the assumption that the warnings should only open if there er no errors. That's why I didn't implement the sorting, and solved it with a `if` case.

schnittstabil: > I had the assumption that the warnings should only open if there er no errors.

Aww sorry, yes that make sense. Nevertheless, I think sorting _additionally_, would be reasonable. Would you mind adding it? No problem, if you don't, I could take care of it too.

tdeschryver: No problem, I'll add it later today!

tdeschryver: @schnittstabil Some feedback would be appreciated :)

For 0, I think we can assume the counts are always numbers (and should they be `undefined` they will be filtered out). See [index.L10](https://github.com/sindresorhus/xo/blob/master/index.js#L10) and [index.L68](https://github.com/sindresorhus/xo/blob/master/index.js#L68)
Are my assumptions good? Should the warnings only be opened if they are no erros?

I would prefer using the descending sorter, that way the file with the most errors will be opened first.

Thoughts?

schnittstabil: > For 0, I think we can assume the counts are always numbers

:+1: I just wasn't sure about it.

> Should the warnings only be opened if there are no errors?

Both make sense, but I've no strong preference. Feel free to decide it.

> I would prefer using the descending sorter, that way the file with the most errors will be opened first.

Mhm, I'm using sublime: all files are opend, and the last file gets the focus. What editor do you use?

tdeschryver: Yea you are right, it focuses the last opened file. Didn't thought about it...

@sindresorhus Should the warnings only be opened if there are no errors?

main.js
```diff
@@ -77,20 +77,21 @@ const log = report => {
...
+		line: result.messages[0].line,
+		column: result.messages[0].column
+	}));
+
 const open = report => {
-	if (report.errorCount === 0) {
-		return;
+	if (report.errorCount > 0) {
+		openEditor(files(report, result => result.errorCount > 0));
+	}	else if (report.warningCount > 0) {
```
schnittstabil: Please replace the `\t` by a space character:

```diff
-}	else if (report.warningCount > 0) {
+} else if (report.warningCount > 0) {
```

tdeschryver:
@schnittstabil I was thinking about putting the `open` under tests too, but I don't know how to test `openEditor` + I didn't want to expose the `files` and `open` functions just so it could be tested.
I do think your suggestions are making sense for future use.

Could you give some feedback about what has to be done?
 -  move `files` and `open` to `open-report.js`
    - but what about `openEditor`? Imo that shouldn't be tested and if it does, how do I test it?  
    - what I am suggesting is that we have a `filesToBeOpened`(?) in `open-report.js` and in `main.js` we would do something like `const open = report => filesToBeOpened(report)`. `filesToBeOpened` would have some tests, but `open` not in my proposition.
    
 - extract the sorting so it can be used later on to sort the messages + test the sorting method
 

Thanks for the reviews!

main.js
Qix-: @sindresorhus How is this allowed by XO?

schnittstabil:
> I didn't want to expose the `files` and `open`

`open-report.js` would not be documented (similar to `gitignore.js` and `options-manager.js`), thus it wouldn't be part of the official XO API, and we therefore can expose whatever we want.

However, I would suggest a simple approach, similar to:

```js
// open-report.js
const openEditor = require('open-editor');

const files = //…

const openReport = (report, opts) => {
	opts = Object.assign({
		openEditor
	}, opts);

	// …
	opts.openEditor(files(…));
	// …
};

module.exports = openReport;
```

```js
// main.js
const openReport = require('./lib/open-report');
// …
if (opts.open) {
	openReport(report);
}
```

```js
// test/open-report.js
const openReport = require('./lib/open-report');

test('…', t => {
	t.plan(1);

	openReport(report, {openEditor: files => {
		t.is(…);
	});
});
```

tdeschryver:
@schnittstabil I used the `proxyquire` lib as in the tests of `options-manager` to stub `openReport`.
Could you give it another review?

lib/open-report.js
```diff
@@ -0,0 +1,28 @@
...
+	if (report.errorCount > 0) {
+		return files(report, result => result.errorCount > 0);
+	} else if (report.warningCount > 0) {
+		return files(report, result => result.warningCount > 0);
+	}
+
+	return [];
+};
+
+module.exports = report => openEditor(filesToOpen(report));
```
schnittstabil: Please, inline the `filesToOpen`, i.e.:
```js
module.exports = report => {
	if (report.errorCount > 0) {
		openEditor(files(report, result => result.errorCount > 0));
	} else if (report.warningCount > 0) {
		openEditor(files(report, result => result.warningCount > 0));
	}
};
```

lib/open-report.js
```diff
@@ -0,0 +1,28 @@
...
+module.exports.openEditor = openEditor;
```
schnittstabil: I see no need to export `openEditor`, please remove it.

main.js
```diff
@@ -4,45 +4,45 @@ const updateNotifier = require('update-notifier');
...
-	  $ xo --space
-	  $ xo --env=node --env=mocha
-	  $ xo --init --space
-	  $ xo --plugin=react
-	  $ xo --plugin=html --extension=html
-	  $ echo 'const x=true' | xo --stdin --fix
-
-	Tips
-	  Put options in package.json instead of using flags so other tools can read it.
+  Usage
```
schnittstabil: Please revert these `\t` to space changes.

main.js
```diff
@@ -65,8 +65,8 @@ const input = cli.input;
 const opts = cli.flags;
 
 const log = report => {
-	// Legacy
-	// TODO: Remove in 1.0.0
+  // Legacy
+  // TODO: Remove in 1.0.0
```
schnittstabil: I do not understand why XO does not complain about that change :confused: – Please revert it.

test/open-report.js
```diff
@@ -0,0 +1,77 @@
...
+	const report = createReport({
+		errorCount: 1,
+		warningCount: 2,
+		results: [
+			createResult({errorCount: 7, filePath: 'seven-errors.js', messages: [createMessage({line: 7, column: 17})]}),
+			createResult({warningCount: 3, filePath: 'three-warnings.js', messages: [createMessage({line: 3, column: 13})]}),
+			createResult({errorCount: 1, filePath: 'one-error.js', messages: [createMessage()]}),
+			createResult({warningCount: 1, filePath: 'one-warning.js', messages: [createMessage()]})
+		]
+	});
```
schnittstabil: IMO, manually creating reports is rather fragile: when ESLint changes the report format, we have to change all those create-functions.

We are already testing on a higher level: we want to assert that the _output_ is correct (i.e. that `open-editor` is called with the right arguments), hence testing with _real input_ is sensible, for example:

```js
const xo = require('..');
// …
const report = xo.lintText(`// TODO xxx
const x =       4;
`, {filename: 'one-warning-and-two-errors.js'});
```

We could also reuse or add files to `test/fixtures` for more complicated tests.

lib/open-report.js
tdeschryver: I had to export it otherwise I couldn't stub it in tests. I'll see what I can do about it.

main.js
tdeschryver: Sorry about those, I just ran `xo --fix` and thought it would fix all of these.

lib/open-report.js
tdeschryver: Seems like I jumped to conclusion too fast, the export isn't needed.

test/open-report.js
tdeschryver: Indeed that's a better approach. I do think we need to use fixtures because we're going to lint multiple files. Should I create new fixtures for this?

tdeschryver: I've pushed my changes using fixtures. 
If its easier for you to make the changes instead of reviewing it after every change, you can go ahead (not that I don't want to do it, in fact I'm glad with your feedback!).

schnittstabil:
Looks almost good to me. I would like to _squash and merge_ this, but you've added the #222 commit too, thus I'm not sure if this could mess the git history. Would you mind in rebasing your PR on the current `master`, to get rid of that commit?

tdeschryver:
@schnittstabil is this ok?

schnittstabil:
:+1: LGTM–sorry for the delay.

@sindresorhus @jamestalmage @marionebl: do you see further improvements?

sindresorhus:
You have to also sort the `messages` in a result object.

For example, this result object is opening the warning as that's the first message:

```js
{ filePath: '/Users/sindresorhus/dev/oss/ava/lib/cli.js',
  messages:
   [ { ruleId: 'complexity',
       severity: 1,
       message: 'Arrow function has a complexity of 22.',
       line: 23,
       column: 15,
       nodeType: 'ArrowFunctionExpression',
       source: 'exports.run = () => {' },
     { ruleId: 'indent',
       severity: 2,
       message: 'Expected indentation of 3 tabs but found 2.',
       line: 60,
       column: 1,
       nodeType: 'Identifier',
       source: '\t\tstring: [',
       endLine: 60,
       endColumn: 3,
       fix: [Object] } ]
  errorCount: 42,
  warningCount: 1,
  ...
}
```

tdeschryver:
My bad, didn't think of that case. Should be solved now.

sindresorhus:
Thank you @tdeschryver. Really nice work on this :)

tdeschryver:
Thanks for the help @schnittstabil!

schnittstabil:
Many thanks for your persistence :blush:

PR #228, opened 2017-06-11
updated package.json

EsrefDurna:
updated package arrify from 1.0.0 to 1.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "simple"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


Qix-:
This is unnecessary. Patch bumps automatically get downloaded. No need to update it in the package.json unless it's a minor or major build.

sindresorhus:
https://nodesource.com/blog/semver-tilde-and-caret

PR #237, opened 2017-06-23
Don't glob everything when XO receives no patterns

sindresorhus:
Currently if you just run `$ xo`, it globs all files and then filters out the supported files. We should instead create the default pattern based on the `extensions` option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,17 +84,19 @@ exports.lintText = (str, opts) => {
 	return processReport(report, opts);
 };
 
-exports.lintFiles = (inputPatterns, opts) => {
+exports.lintFiles = (patterns, opts) => {
 	opts = optionsManager.preprocess(opts);
 
+	const isEmptyPatterns = patterns.length === 0;
 	const defaultPattern = `**/*.{${opts.extensions.join(',')}}`;
-	const patterns = inputPatterns.length === 0 ? [defaultPattern] : arrify(inputPatterns);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -84,23 +84,24 @@ exports.lintText = (str, opts) => {
 	return processReport(report, opts);
 };
 
-exports.lintFiles = (patterns, opts) => {
+exports.lintFiles = (inputPatterns, opts) => {
 	opts = optionsManager.preprocess(opts);
-	patterns = patterns.length === 0 ? ['**/*'] : arrify(patterns);
 
+	const defaultPattern = `**/*.{${opts.extensions.join(',')}}`;
+	const patterns = inputPatterns.length === 0 ? [defaultPattern] : arrify(inputPatterns);
```
schnittstabil: Using two very similar variables/constants (`inputPatterns` and `patterns`)  within the same scope makes it harder to read and maintain IMO. As far as I can see, below your line 91, `inputPatterns.length` is the only important information, therefore we could use a constant like `patternsAreEmpty` instead:

```js
exports.lintFiles = (patterns, opts) => {
	//…

	const patternsAreEmpty = patterns.length === 0;
	patterns = patternsAreEmpty ? [defaultPattern] : arrify(patterns);

	//…

	return globby(patterns, …

		//…
		if (!patternsAreEmpty) {
```

sindresorhus: Good point. Fixed.

PR #246, opened 2017-08-15
Update outdated note about config overrides

not-an-aardvark:
Hello :wave:

Starting in ESLint v4.1.0, ESLint now supports [config overrides based on glob patterns](http://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) (with a design largely inspired by xo, I think). This commit updates the readme to remove the note about ESLint not supporting config overrides.


kevva:
Thanks :).

PR #251, opened 2017-09-11
Use `lodash.mergewith` instead of `deep-assign`

### LINKED ISSUE #241: Switch from `deep-assign` to `lodash.merge`
Since `deep-assign` is deprecated. Note that it might not be just a clean switch.

kevva:
Fixes #241.

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -2,7 +2,7 @@
 const os = require('os');
 const path = require('path');
 const arrify = require('arrify');
-const deepAssign = require('deep-assign');
+const mergeWith = require('lodash.mergewith');
 const multimatch = require('multimatch');
 const pathExists = require('path-exists');
 const pkgConf = require('pkg-conf');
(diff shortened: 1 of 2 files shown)


lib/options-manager.js
```diff
@@ -199,8 +207,8 @@ const findApplicableOverrides = (path, overrides) => {
 
 const mergeApplicableOverrides = (baseOptions, applicableOverrides) => {
 	applicableOverrides = applicableOverrides.map(normalizeOpts);
-	const overrides = [emptyOptions(), baseOptions].concat(applicableOverrides);
-	return deepAssign.apply(null, overrides);
+	const overrides = [emptyOptions(), baseOptions].concat(applicableOverrides, mergeFn);
+	return mergeWith.apply(null, overrides);
```
kevva: Any reason why we aren't returning `mergeWith(emptyOptions(), baseOptions, applicableOverrides, mergeFn`?

kevva: Oh, because `applicableOverrides` is an `Array`.

PR #254, opened 2017-09-19
Adds be-polite rule from no-shit plugin

jakubsadura:
Let's make sure we're all polite


kevva:
While it's nice addition and a good idea, I don't think it should be included in `xo` by default since it might decrease performance further which is a problem right now. And let's be honest, you shouldn't need a linter to tell you not to use that kind of language in code.

Issue #266, opened 2017-11-15
Enforce rules based on "engines" field in package.json


sindresorhus:
So if for example, Node.js 8 is targeted we could enable some useful Node.js 8+ rules.

I've been wanting to enable some rules, but I don't want to introduce yet another option for it and be forced to use it in every repo. 

If there's no `"engines"` field in the project package.json, we will just default to how it works now.

It should also be possible to turn this off with an option for use in other environments.

Some things I'd like to enable:

- https://github.com/sindresorhus/eslint-config-xo/blob/891eb4c9bd1e33af8a12fd6f64d0629c08b88785/esnext.js#L15-L24
-
 https://github.com/sindresorhus/xo/blob/8ac0f3e25426937ef96288dfbe7d4dfda857c9bf/config/plugins.js#L33-L34
- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/115
- And probably more in the future.

sindresorhus:
> About situations where engines.node is not defined in package.json, shouldn't xo produce a warning or even error if it's omitted?

Warning

> xo could add a backports: true setting for people who want new rules to be enabled by semver-minor releases of xo. The backports rules could be declared in a way that they can easily be switched on by default during semver-major release.

Nah. I think that's overcomplicating it.

> A similar check that might be nice is no-legacy-backport-modules. For example directly depending on or requiring the object-assign module with engines.node >= 4 is likely pointless.

Sounds like a potential rule for https://github.com/mysticatea/eslint-plugin-node

> Eventually I'd like to see xo able to report an error if a module has dependencies that require a higher version of node. Say I have a module that requires node >=0.10 and run ncu -u to update all dependencies. This updates pify from 2.3.0 to 3.0.0 which is an unsupported module for node <4.

Sounds like a useful rule for https://github.com/mysticatea/eslint-plugin-node or https://github.com/benmosher/eslint-plugin-import You should open an issue about it there.

coreyfarrell:
For now I've opened npm/node-semver#228.  Any rule comparing the `engines.node` range between multiple packages will require this.

PR #268, opened 2017-11-22
Add eslint-plugin-node

### LINKED ISSUE #123: eslint-plugin-node
It has some interesting rules, some of which we've already discussed, and some that are already covered, but I think there is stuff we can use there:

https://github.com/mysticatea/eslint-plugin-node

pvdlg:
Fixes #123

Add the following rules:

| Rule ID                                                                                                                     | Description                                                       |
| --------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------- |
| [no-unpublished-bin](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-bin.md)         | Disallow 'bin' files which are ignored by npm                     |
| [no-unpublished-import](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-import.md)   | Disallow `import` declarations of private things                  |
| [no-unpublished-require](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md) | Disallow `require()` expressions of private things                |
| [process-exit-as-throw](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/process-exit-as-throw.md)   | Make `process.exit()` expressions the same code path as `throw`   |
| [no-deprecated-api](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md)     | Disallow unsupported ECMAScript features on the specified version |
| [exports-style](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/exports-style.md)                   | Enforce either module.exports or exports                          |

The other rules provided are disabled for the following reasons:

| Rule ID                                                                                                                       | Reason for disabling                                                                  |
| ----------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------- |
| [no-extraneous-import](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-import.md)       | Redundant with `import/no-extraneous-dependencies`                                    |
| [no-extraneous-require](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md)     | Redundant with `import/no-extraneous-dependencies`                                    |
| [no-missing-import](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-missing-import.md)             | Redundant with `import/no-unresolved`                                                 |
| [no-missing-require](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-missing-require.md)           | Redundant with `import/no-unresolved`                                                 |
| [no-unsupported-features](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md) | Doesn't allow to exclude compiled sources                                             |
| [shebang](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/shebang.md)                                 | Doesn't exclude scripts executed with node but not referenced in "bin" (i.e main.js)  |
| ~~[exports-style](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/exports-style.md)~~                 | ~~Too restrictive, forces to use either `modules.exports` or `exports` but not both~~ |

Let me know if I should make changes in the rules enabled/disabled or in the configuration.

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -78,7 +78,7 @@ module.exports = {
 		// Disabled as the rule doesn't allow to exclude compiled sources
 		// 'node/no-unsupported-features': 'error',
 		'node/process-exit-as-throw': 'error',
-		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin" (i.e main.js)
+		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin". See https://github.com/mysticatea/eslint-plugin-node/issues/96
 		// 'node/shebang': 'error',
 		'node/no-deprecated-api': 'error',
 		'node/exports-style': ['error', 'module.exports']
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -15,7 +15,8 @@ module.exports = {
 		'ava',
 		'unicorn',
 		'promise',
-		'import'
+		'import',
+		'node'
 	],
 	extends: [
 		'plugin:ava/recommended',
@@ -64,6 +65,22 @@ module.exports = {
 		'import/prefer-default-export': 'error',
 		'import/no-unassigned-import': ['error', {
 			allow: ['babel-polyfill', '@babel/polyfill', 'babel-register', '@babel/register']
-		}]
+		}],
+		// Redundant with import/no-extraneous-dependencies
+		// 'node/no-extraneous-import': 'error',
+		// 'node/no-extraneous-require': 'error',
+		// Redundant with import/no-unresolved
+		// 'node/no-missing-import': 'error',
+		// 'node/no-missing-require': 'error',
+		'node/no-unpublished-bin': 'error',
+		'node/no-unpublished-import': ['error', {allowModules: ['electron', 'atom']}],
+		'node/no-unpublished-require': ['error', {allowModules: ['electron', 'atom']}],
+		// Disabled as the rule doesn't allow to exclude compiled sources
+		// 'node/no-unsupported-features': 'error',
+		'node/process-exit-as-throw': 'error',
+		// Disable as the rule doesn't exclude scripts executed with node but not referenced in "bin" (i.e main.js)
+		// 'node/shebang': 'error',
+		'node/no-deprecated-api': 'error',
+		'node/exports-style': ['error', 'module.exports', {allowBatchAssign: true}]
 	}
 };
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -54,7 +54,7 @@ module.exports = {
 		// 'import/unambiguous': 'error',
 		// enable this sometime in the future when Node.js has ES2015 module support
 		// 'import/no-commonjs': 'error',
-		// looks useful, but too unstable at the moment
+		// Looks useful, but too unstable at the moment
 		// 'import/no-deprecated': 'error',
 		'import/no-extraneous-dependencies': 'error',
 		'import/no-mutable-exports': 'error',
@@ -78,7 +78,7 @@ module.exports = {
 		// Disabled as the rule doesn't allow to exclude compiled sources
 		// 'node/no-unsupported-features': 'error',
 		'node/process-exit-as-throw': 'error',
-		// Disable as the rule doesn't exclude scripts executed with node but not referenced in "bin" (i.e main.js)
+		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin" (i.e main.js)
 		// 'node/shebang': 'error',
 		'node/no-deprecated-api': 'error',
 		'node/exports-style': ['error', 'module.exports', {allowBatchAssign: true}]
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -81,6 +81,6 @@ module.exports = {
 		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin" (i.e main.js)
 		// 'node/shebang': 'error',
 		'node/no-deprecated-api': 'error',
-		'node/exports-style': ['error', 'module.exports', {allowBatchAssign: true}]
+		'node/exports-style': ['error', 'module.exports']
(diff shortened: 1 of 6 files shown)


sindresorhus:
>  no-unsupported-features | Doesn't allow to exclude compiled sources
-- | --

Can you open an issue on `eslint-plugin-node` about supporting this?

> shebang | Doesn't exclude scripts executed with node but not referenced in "bin" (i.e main.js)
-- | --

This too.

> exports-style | Too restrictive, forces to use either modules.exports or exports but not both
-- | --

And this.

config/plugins.js
```diff
@@ -64,6 +65,23 @@ module.exports = {
...
-		}]
+		}],
+		// Redundant with import/no-extraneous-dependencies
+		// 'node/no-extraneous-import': 'error',
+		// 'node/no-extraneous-require': 'error',
+		// Redundant with import/no-unresolved
+		// 'node/no-missing-import': 'error',
+		// 'node/no-missing-require': 'error',
+		'node/no-unpublished-bin': 'error',
+		'node/no-unpublished-import': 'error',
```
sindresorhus: Include `electron` and `atom` in the `allowModules` option: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-import.md#allowmodules For ease of use with Electron and Atom.

config/plugins.js
```diff
@@ -64,6 +65,23 @@ module.exports = {
...
+		'node/no-unpublished-require': 'error',
```
sindresorhus: Same as above.

pvdlg:
> no-unsupported-features | Doesn't allow to exclude compiled sources

mysticatea/eslint-plugin-node#97

> shebang | Doesn't exclude scripts executed with node but not referenced in "bin" (i.e main.js)

mysticatea/eslint-plugin-node#96

> exports-style | Too restrictive, forces to use either modules.exports or exports but not both

I first though it was possible to use both `modules.exports` and `exports` in the same project as long as it's consistent in the same js file. After more tests it turns out I was wrong.
Using export style that are consistent within each file but inconsistent across the project does create the issue described in the rule.
So I went ahead and enabled the rule

config/plugins.js
pvdlg: Done!

config/plugins.js
pvdlg: Done!

pvdlg:
@sindresorhus per this [comment](https://github.com/mysticatea/eslint-plugin-node/issues/97#issuecomment-347140801) we could use the override feature to disable those rules for certain files.

In `xo` we could by default enable the `shebang` rule only for the files in the `bin` property of the `package.json`.

Regarding `no-unsupported-features` we could enable it only for certain files, but I don't know if there is a good default setting. We can either:
- keep it disabled and let users enable it if they want to
- Enable it and exclude ava files by default and let the users configure the `overrides` property to exclude more
- Add a new property in `xo` called something like `compiledSources` that will allow to define an Array of globs. And we could use it to disable certain rules for files that match.

What's your thoughts ?
Also I'm not sure that should be part of this PR or if it should be implemented in a following one.

sindresorhus:
> In xo we could by default enable the shebang rule only for the files in the bin property of the package.json.

Yes, let's wait an see the outcome of https://github.com/mysticatea/eslint-plugin-node/issues/96

> Regarding no-unsupported-features we could enable it only for certain files, but I don't know if there is a good default setting. We can either:

Let's keep it disabled. I don't want lots of complication for something that is not super useful with latest Node.js, as there are not that many commonly used unsupported features.

config/plugins.js
```diff
@@ -64,6 +65,22 @@ module.exports = {
...
+		// Redundant with import/no-unresolved
+		// 'node/no-missing-import': 'error',
+		// 'node/no-missing-require': 'error',
+		'node/no-unpublished-bin': 'error',
+		'node/no-unpublished-import': ['error', {allowModules: ['electron', 'atom']}],
+		'node/no-unpublished-require': ['error', {allowModules: ['electron', 'atom']}],
+		// Disabled as the rule doesn't allow to exclude compiled sources
+		// 'node/no-unsupported-features': 'error',
+		'node/process-exit-as-throw': 'error',
+		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin" (i.e main.js)
```
sindresorhus: Can you link to https://github.com/mysticatea/eslint-plugin-node/issues/96 ?

config/plugins.js
```diff
@@ -64,6 +65,22 @@ module.exports = {
...
+		// 'node/shebang': 'error',
+		'node/no-deprecated-api': 'error',
+		'node/exports-style': ['error', 'module.exports', {allowBatchAssign: true}]
```
sindresorhus: I don't think `allowBatchAssign` should be `true`. It's not a good pattern.

config/plugins.js
pvdlg: Done

config/plugins.js
pvdlg: Done

pvdlg:
> Let's keep it disabled. I don't want lots of complication for something that is not super useful with latest Node.js, as there are not that many commonly used unsupported features.

This is something we can implement within the scope of #266

sindresorhus:
> This is something we can implement within the scope of #266

Agreed :)

sindresorhus:
Great! Thank you 🙌

Would you be interested in joining the project? You done some high-quality contributions :)

pvdlg:
I'll take a stab at #266 sometime (in a weeks I guess).

pvdlg:
> Would you be interested in joining the project? You done some high-quality contributions :)

Absolutely !

PR #269, opened 2017-11-24
Exclude babel-register and babel-polyfill from no-unassigned-import

pvdlg:
[babel-polyfill](https://babeljs.io/docs/usage/polyfill/) and [babel-register](https://babeljs.io/docs/usage/babel-register/) are common modules that doesn't need to assigned.

It would be convenient to have them ignored by the [import/no-unassigned-import](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md) rule.


sindresorhus:
Good idea :)

PR #271, opened 2017-11-25
Add optional Prettier support

### LINKED ISSUE #176: Integrate the `prettier` module
would you consider using or doing something like https://github.com/kentcdodds/prettier-eslint for xo --fix ?

more links 
https://github.com/jlongster/prettier
https://github.com/kentcdodds/prettier-eslint-cli

pvdlg:
Add the `prettier` boolean options. The default value is `false`.

If `prettier` options is set to `true`:
- `xo --fix` will format the code with [Prettier](https://github.com/prettier/prettier)
- `xo` will report errors for each snippet of code that is not formatted in accordance with Prettier format

Under the hood it uses:
- [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to overrides the ESlint stylistic rules that are covered by Prettier
- [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) to format with `--fix`  and to report errors when the code is not formatted in accordance with Prettier

The `prettier` dependency is explicitly set with `~` as some small format change can occur on minor `prettier` releases. This way no unexpected change will happen for `xo` users. They will see a potential change in formatting only after the `xo` team upgrade the `prettier` dependency version and releases a new `xo` version.
That said, in my ~6 months of using Prettier, it happened only once that a minor version upgrade required me to change my code formatting.

Fix #176


sindresorhus:
The XO `space` and `semicolon` options should apply to Prettier too.

pvdlg:
> The XO space and semicolon options should apply to Prettier too.

Yes! I'm on it :)

pvdlg:
I pushed the change. Let me know if anything else is needed!

sindresorhus:
Tried it on a big codebase now and realized we need to set the Prettier `singleQuote` option to `true`, `useTabs` to `true` if the XO `space` option is `false`, and `bracketSpacing` to `false`.

main.js
```diff
@@ -20,6 +20,7 @@ const cli = meow(`
 	  --ignore          Additional paths to ignore  [Can be set multiple times]
 	  --space           Use space indent instead of tabs  [Default: 2]
 	  --no-semicolon    Prevent use of semicolons
+	  --prettier        Format code with Prettier
```
sindresorhus: Also add it to the `flags` option: https://github.com/sindresorhus/xo/blob/bc66114d3381e24b4272dfd88673643a9cff249e/main.js#L47

sindresorhus: And should we maybe clarify that the flag itself doesn't format the code but should be used with the `--fix` flag?

lib/options-manager.js
```diff
@@ -179,6 +179,23 @@ const buildConfig = opts => {
...
+	if (opts.prettier) {
+		// The prettier plugin uses Prettier to format the code with `--fix`
+		config.plugins = config.plugins.concat('prettier');
+		// The prettier config overrides ESLint stylistic rules that are handled by Prettier
+		config.baseConfig.extends = config.baseConfig.extends.concat('prettier');
+		// The `prettier/prettier` rule reports errors if the code is not formatted in accordance to Prettier
+		config.rules['prettier/prettier'] = ['error', {tabWidth: spaces, semi: opts.semicolon !== false}];
+		// If the user has the React, Flowtype or Standard plugin, add the corresponding Prettier rule overrides
+		// See https://github.com/prettier/eslint-config-prettier for the list of plugins overrrides
+		['react', 'flowtype', 'standard'].forEach(override => {
```
sindresorhus: Use a `for-of` loop.

sindresorhus:
Can you also mention Prettier support in the feature highlight section: https://github.com/sindresorhus/xo#highlights

sindresorhus:
When I run this PR on my [AVA](https://github.com/avajs/ava) project I get some weird output:

```
  lib/snapshot-manager.js:89:34
  ✖   89:34  Replace A·||·buffer[newLength]·===·0x0D with a·||·buffer[newLength]·===·0x0d  prettier/prettier
  ✖  242:42  Replace A with a                                                              prettier/prettier

  test/cli.js:843:45
  ✖  843:45  Replace A with a                                                              prettier/prettier
  ✖  864:45  Replace A,·0xFF,·0xFF with a,·0xff,·0xff                                      prettier/prettier
  ✖  885:45  Replace A with a                                                              prettier/prettier
```

pvdlg:
If we do so, the config we pass to `eslint-plugin-prettier` will override the Prettier configuration the user might have. See https://prettier.io/docs/en/configuration.html.

So for user who currently use ESLint with Prettier with for example `bracketSpacing` to `true` and wants to migrate to `xo`, that's going to be a problem as this will completely change their codebase.

I see 2 solutions.

1. XO config for what's configurable by the user / Prettier conf for the rest 

`tabWidth` and `semi` are set based on XO options `space` and `semicolon`. We should add `useTabs` to that.
The other options will be defined by Prettier (we can mention that in the README).
 
2. Read Prettier config and sets XO default if not set in Prettier

We would have to retrieve the Prettier config with `cosmiconfig` and determine the config of each property in this order :
- `semi`, `tabWidth` and `useTabs` based on XO `space` and `semicolon` (wether they are set explicitly or not)
-`singleQuote`, `trailingComma` and `bracketSpacing` based on Prettier config is set, otherwise XO default: `singleQuote: true`, `trailingComma: none`, `bracketSpacing: false`

pvdlg:
> When I run this PR on my AVA project I get some weird output:

What do you mean by "weird"? It's the pretty plugin that report errors because your code doesn't fit the Prettier format.

main.js
pvdlg: I clarified the option description with `Conform to Prettier code style` as it does two things:
- report errors if the code is not formatted according to Prettier format
- automatically format with --fix (as with any other rules)

Let me know if that's clear enough.

I also added the flag

lib/options-manager.js
pvdlg: Done

pvdlg:
> Can you also mention Prettier support in the feature highlight section: https://github.com/sindresorhus/xo#highlights

Done

pvdlg:
@sindresorhus what's you though on [comment](https://github.com/sindresorhus/xo/pull/271#issuecomment-347579525) ?

Option 2 is the least disruptive for both
- current `xo` users who activate the `prettier` options,  as it will keep the same style
- current `prettier` users who switch to `xo`, as their `prettier` config will be maintained

Regarding option 3, Prettier tries to add as few options as possible and do so only when there is a huge amount of demands from the community. So if Prettier added options for `quote`, `trailingComma` and `bracketSpacing` its probably because a lot of people want that. So it might make sense to add them to `xo`.

pvdlg:
I went ahead with solution 2. So now XO set default values for `singleQuote`, `trailingComma`, `bracketSpacing` and `jsxBracketSameLine` unless they are defined in `prettier` config.

If we decide to add more options to XO (options 3) that would be easy to modify

readme.md
```diff
@@ -32,6 +32,7 @@ Uses [ESLint](http://eslint.org) underneath, so issues regarding rules should be
 - Fix many issues automagically with `$ xo --fix`.
 - Open all files with errors at the correct line in your editor with `$ xo --open`.
 - Specify [indent](#indent) and [semicolon](#semicolon) preferences easily without messing with the rule config.
+- Optionnaly use the [Prettier](https://github.com/prettier/prettier) code style.
```
bramkok: Optionally

pvdlg: Fixed. Thanks!

sasha240100:
@pvdlg @sindresorhus What's the current status of this PR?

pvdlg:
On my side it's ready to fully reviewed and merged

package.json
```diff
@@ -92,6 +94,7 @@
 		"open-editor": "^1.0.1",
 		"path-exists": "^3.0.0",
 		"pkg-conf": "^2.0.0",
+		"prettier": "~1.8.2",
```
alrik: @pvdlg The latest prettier version is currently `1.9.2`.

pvdlg: Thanks. It's updated!

sindresorhus:
This looks great @pvdlg. Thanks for working on it! :)

Merry xmas and happy new year ✨

pvdlg:
Thanks! Merry xmas and happy new year to you as well ✨

sholladay:
So if I understand this correctly, using `--prettier` will disable all style rules from XO.

Seems like a better approach would be what [prettier-eslint](https://github.com/prettier/prettier-eslint) does. That is to say, it runs `prettier` first and then passes the output to `eslint --fix`. That way the code has consistency (which `prettier` is better at) while still being opinionated (which `eslint` / `xo` is better at).

azz:
This is awesome. We've just recently released `1.10` (https://prettier.io/blog/2018/01/10/1.10.0.html), we should update the version number here too! (I can PR this?)

PR #275, opened 2017-12-18
Added lockfile

### LINKED ISSUE #274: Add package.lock or similar
This morning my builds are broken because a dependency of yours released a new version, my runner took it and xo doesn't work anymore.

I manually added to my package.json an old version of the dependency.

Please see https://github.com/mafintosh/is-my-json-valid/issues/149

kutekat:
Fixes #274

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5967 @@
+{
+	"name": "xo",
+	"version": "0.18.2",
+	"lockfileVersion": 1,
+	"requires": true,
+	"dependencies": {
+		"@ava/babel-plugin-throws-helper": {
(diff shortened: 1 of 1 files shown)


LinusU:
Actually, this doesn't fix the issue since `package-lock.json` is automatically filtered out by `npm pack` (and in effect `npm publish`). There will be no difference to the published package after this patch is applied.

The correct approach, in my opinion, would be for each _end user_ to add their own `package-lock.json`-file, thus ensuring that they always get the expected dependencies.

kutekat:
Thank you for reply

PR #279, opened 2018-01-06
Discover lint options relative to input location

### LINKED ISSUE #223: Overrides not working when calling xo with absolute paths
I have this in my package.json:

```json
"xo": {
  "extends": "xo-react",
  "parser": "babel-eslint",
  "plugins": ["react"],
  "envs": ["node", "browser"],
  "rules": {
    "react/forbid-component-props": 0,
    "react/jsx-closing-bracket-location": ["error", "tag-aligned"]
  },
  "overrides": [{
    "files": "src/client/components/*.js",
    "rules": {
      "unicorn/filename-case": [
        "error", {"case": "pascalCase"}
      ]
    }
  }, {
    "files": "src/client/containers/*.js",
    "rules": {
      "unicorn/filename-case": [
        "error",
        {"case": "pascalCase"}
      ]
    }
  }]
}
```

And I have the file `src/client/components/ExtendedReferenceManyField.js`

Now, if I call

```
$ xo src/client/components/ExtendedReferenceManyField.js
```

it finishes without errors. However, calling:

```
$ xo /home/user/dev/project/src/client/components/ExtendedReferenceManyField.js
```

will output:

```
  src/client/components/ExtendedReferenceManyField.js:1:1
  ✖  1:1  Filename is not in kebab case. Rename it to extended-reference-many-field.js.  unicorn/filename-case

  1 error
```

I stumbled on this while using [lint-staged](https://github.com/okonet/lint-staged) in my project.

---

**xo version**: 0.18.2
**Node version**: 6.10.1
**OS**: Linux Mint 18 Cinnamon 64-bit

zanona:
This commit allows running commands from any location and still have
`xo` to analyse files against the input directory rather than where the
command is initiated (`process.cwd`)

Previously commands like:

```bash
cd /Users/foo/Desktop
xo /Users/foo/Develop/example.com/lib/index.js
```

or

```bash
cd /Users/foo/Desktop
FILE=/Users/foo/Develop/example.com/lib/index.js
cat $FILE | xo --stdin --stdin-filename $FILE
```

The above would lookup for `xo` options under `package.json` relative to
`/Users/foo/Desktop` rather than where the file is located, which is not
the intended behaviour.

This commit tries to solve that always analyse against the command
input location instead, whener available.

This has benefits such as solving issues like
https://github.com/sindresorhus/xo/issues/223 and
https://github.com/okonet/lint-staged/issues/369
  
---

Fixes #223

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,11 +35,15 @@ const runEslint = (paths, opts) => {
 	const config = optionsManager.buildConfig(opts);
 	const engine = new eslint.CLIEngine(config);
 	const report = engine.executeOnFiles(paths, config);
-
 	return processReport(report, opts);
 };
 
-module.exports.lintText = (str, opts) => {
+module.exports.lintText = (str, opts = {}) => {
+	if (opts.stdinFilename && !opts.cwd) {
+		opts.cwd = path.dirname(opts.stdinFilename);
+		opts.filename = opts.stdinFilename;
+	}
+
 	opts = optionsManager.preprocess(opts);
 
 	if (opts.overrides && opts.overrides.length > 0) {
@@ -84,10 +88,17 @@ module.exports.lintText = (str, opts) => {
 	return processReport(report, opts);
 };
 
-module.exports.lintFiles = (patterns, opts) => {
+module.exports.lintFiles = (patterns, opts = {}) => {
+	const isEmptyPatterns = patterns.length === 0;
+
+	if (!isEmptyPatterns && !opts.cwd) {
+		// Use file path rather process.cwd for analysing files
+		const cwd = path.dirname(arrify(patterns)[0]);
+		opts.cwd = cwd;
+	}
+
 	opts = optionsManager.preprocess(opts);
 
-	const isEmptyPatterns = patterns.length === 0;
 	const defaultPattern = `**/*.{${opts.extensions.join(',')}}`;
 	const ignoreFilter = optionsManager.getGitIgnoreFilter(opts);
 
--- a/index.js
+++ b/index.js
@@ -35,10 +35,12 @@ const runEslint = (paths, opts) => {
 	const config = optionsManager.buildConfig(opts);
 	const engine = new eslint.CLIEngine(config);
 	const report = engine.executeOnFiles(paths, config);
+
 	return processReport(report, opts);
 };
 
-module.exports.lintText = (str, opts = {}) => {
+module.exports.lintText = (str, opts) => {
+	opts = opts || {};
 	if (opts.stdinFilename && !opts.cwd) {
 		opts.cwd = path.dirname(opts.stdinFilename);
 		opts.filename = opts.stdinFilename;
@@ -88,9 +90,10 @@ module.exports.lintText = (str, opts = {}) => {
 	return processReport(report, opts);
 };
 
-module.exports.lintFiles = (patterns, opts = {}) => {
+module.exports.lintFiles = (patterns, opts) => {
 	const isEmptyPatterns = patterns.length === 0;
 
+	opts = opts || {};
 	if (!isEmptyPatterns && !opts.cwd) {
 		// Use file path rather process.cwd for analysing files
 		const cwd = path.dirname(arrify(patterns)[0]);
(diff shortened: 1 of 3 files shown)


pvdlg:
The implementation for `lintText` looks good. However `lintFiles` is problematic:
- Using the first entry in the patterns is very arbitrary
- What you are going to get in `patterns` are globs not filename, so `path.dirname` will not work  

In the case of `lintText` we should get the config for each files match by `globby`, because each can be related to a different `package.json`. It's especially true in the context of a monorepo.

For example:
```
.
+-- module1/
|   +-- package.json
|   +-- src
|       +-- file1.js
|       +-- file2.js
+-- module2/
|   +-- package.json
|   +-- src
|       +-- file3.js
```
`file1.js` and `file2.js` should be linted with the option of `module1/package.json` and `file2.js` with the options of `module2/package.json`.

This is the way `Prettier` handles its configuration. From the [documentation](https://prettier.io/docs/en/configuration.html):

> The configuration file will be resolved starting from the location of the file being formatted, and searching up the file tree until a config file is (or isn't) found.

In order to be performant we would probably have to maintain a cache so we don't read `module1/package.json` twice to lint `file1.js` and `file2.js`.

That would also require a bit of refactoring in XO as `optionsManager.preprocess` would have to called for each group of file that uses the same `package.json`

We can probably find some implementation ideas in the [prettier config/resolve-config.js](https://github.com/prettier/prettier/blob/master/src/config/resolve-config.js).

index.js
```diff
@@ -85,9 +91,17 @@ module.exports.lintText = (str, opts) => {
 };
 
 module.exports.lintFiles = (patterns, opts) => {
+	const isEmptyPatterns = patterns.length === 0;
+
+	opts = opts || {};
+	if (!isEmptyPatterns && !opts.cwd) {
+		// Use file path rather process.cwd for analysing files
+		const cwd = path.dirname(arrify(patterns)[0]);
```
pvdlg: Using the first entry in the array is arbitrary and it would be inconsistent as the behavior would change depending on the order of the input passed by the user.

In addition is will not work because `patterns` contains globs not file paths. So `path.dirname` will fail.

test/lint-files.js
```diff
@@ -94,3 +94,9 @@ test('multiple negative patterns should act as positive patterns', async t => {
 
 	t.deepEqual(paths, ['!!unicorn.js', '!unicorn.js']);
 });
+
+test('respect file path for option discovery rather than process.cwd', async t => {
+	const filename = path.join(__dirname, 'fixtures', 'overrides', 'index.js');
```
pvdlg: This test is incorrect. What is expected to be passed to `lintFiles` is a list of globs, not a file name.

pvdlg:
To implement the feature the way I describe it in the previous comment we could use [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to find and read the XO config in the `package.json` for each file as it has the advantage of handling a cache (both for files and directory), so the perf should be good.

In addition it would allow to configure XO with an `.xorc` or `xo.config.js` file. Not sure if it's desired though. It could be disabled in `cosmiconfig` if we don't want to allow those type of config files.

pvdlg:
In addition we should also make changes regarding [this line](https://github.com/sindresorhus/xo/blob/master/lib/options-manager.js#L190) as we need to load shareable config relatively to the `package.json` in which they are defined not the current working directory from which XO is executed.

pvdlg:
@sindresorhus would you be open to a PR that implements what I proposed in [this comment](https://github.com/sindresorhus/xo/pull/279#issuecomment-356021126) and [that one](https://github.com/sindresorhus/xo/pull/279#issuecomment-356388248) ?

niftylettuce:
Can we get this in?  Annoying to have to deal with this current bug (overrides and explicit file paths)

Rowno:
Is there anything we can do to help drive this to completion?

sindresorhus:
> would you be open to a PR that implements what I proposed in this comment and that one ?

Yes, let's go with cosmiconfig, but disable YAML.

jorgegonzalez:
Just curious, what would be the point of disabling YAML @sindresorhus?

sindresorhus:
@jorgegonzalez [YAML is not a good format](https://arp242.net/weblog/yaml_probably_not_so_great_after_all.html) and I don't want too many options. I still think having config only in package.json is enough, but I'm willing to budge.

aslafy-z:
#331

PR #281, opened 2018-01-07
Enforce rules based on "engines" field in package.json

### LINKED ISSUE #266: Enforce rules based on "engines" field in package.json

pvdlg:
This allow to enable specific rules/rules configuration based on the range defined in `engines` field in `package.json` or the `nodeVersion` option.

If `nodeVersion` is set to `false` then the old behavior is resumed: no Node version specific rules are enabled.

The following rules are enable:
- `promise/prefer-await-to-then` for Node 8 and above
- `prefer-rest-params` for Node 6 and above
- `prefer-spread` for Node 5 and above
- `prefer-destructuring` for Node 6 and above

Should I add more?

We could later add [node/no-unsupported-features](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md) by passing the value `opts.nodeEngine` in the [version option](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md#version).
Unfortunately for now the `version` option of the rule accept only a specific version and not the range defined in `engines.node` or `nodeEngine`. That should be solved once mysticatea/eslint-plugin-node@846e677 get released. We could enable the rule then.

I also added a "Tips" for overriding the test when using AVA. Maybe we can make that automatic if the users have AVA in their dependencies?

Fix #266

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -8,6 +8,7 @@ const pathExists = require('path-exists');
 const pkgConf = require('pkg-conf');
 const resolveFrom = require('resolve-from');
 const prettier = require('prettier');
+const semver = require('semver');
 
 const DEFAULT_IGNORE = [
 	'**/node_modules/**',
@@ -39,6 +40,43 @@ const DEFAULT_CONFIG = {
 			path.join(__dirname, '../config/overrides.js'),
 			path.join(__dirname, '../config/plugins.js')
 		]
+	}
+};
+
+/**
+ * Define the rules that are enabled only for specific version of Node, based on `engines.node` in package.json or the `node-engine` option.
+ *
+ * The keys ae rule names and the values are an Object with a valid semver (`4.0.0` is valid `4` is not) as keys and the rule configuration as values.
+ * Each entry define the rule configuration and the minimum Node version for which to set it.
+ * The entry with the highest version that is compliant with the `engines.node`/`node-engine` range will be used.
+ *
+ * @type {Object}
+ *
+ * @example
+ * ```javascript
+ * {
+ * 	'plugin/rule': {
+ * 		'6.0.0': ['error', {prop: 'node-6-conf'}],
+ *		'8.0.0': ['error', {prop: 'node-8-conf'}]
+ * 		}
+ * }
+ *```
+ * With `engines.node` set to `>=4` the rule `plugin/rule` will not be used.
+ * With `engines.node` set to `>=6` the rule `plugin/rule` will be used with the config `{prop: 'node-6-conf'}`.
+ * With `engines.node` set to `>=8` the rule `plugin/rule` will be used with the config `{prop: 'node-8-conf'}`.
+ */
+const ENGINE_RULES = {
+	'promise/prefer-await-to-then': {
+		'8.0.0': 'error'
+	},
+	'prefer-rest-params': {
+		'6.0.0': 'error'
+	},
+	'prefer-spread': {
+		'5.0.0': 'error'
+	},
+	'prefer-destructuring': {
+		'6.0.0': ['error', {array: true, object: true}, {enforceForRenamedProperties: true}]
 	}
 };
 
@@ -88,7 +126,8 @@ const mergeWithPkgConf = opts => {
 	opts = Object.assign({cwd: process.cwd()}, opts);
 	opts.cwd = path.resolve(opts.cwd);
 	const conf = pkgConf.sync('xo', {cwd: opts.cwd, skipOnFalse: true});
-	return Object.assign({}, conf, opts);
+	const engines = pkgConf.sync('engines', {cwd: opts.cwd});
+	return Object.assign({}, conf, {engines}, opts);
 };
 
 const normalizeSpaces = opts => {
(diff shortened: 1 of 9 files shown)


sindresorhus:
> Maybe we can make that automatic if the users have AVA in their dependencies?

Yeah, let's do that.

pvdlg:
AVA doesn't seems to expose an API to read its config to figure out the `files` option.

Do I have a convenient way to determine where are the AVA test files for the current user?

If no I can use the default that would be the list [here](https://github.com/avajs/ava/blob/c1faf95b9bbaa1981013e98576fc58a4d6217020/lib/ava-files.js#L82)

pvdlg:
I added the default list of AVA files. Let me know you'd prefer something else.

lib/options-manager.js
```diff
@@ -44,6 +55,43 @@ const DEFAULT_CONFIG = {
...
+	'prefer-rest-params': {
+		'6.0.0': 'error'
+	},
+	'prefer-spread': {
+		'5.0.0': 'error'
+	},
+	'prefer-destructuring': {
+		'6.0.0': ['error', {array: true, object: true}, {enforceForRenamedProperties: true}]
+	}
+};
```
sindresorhus: Can you include https://github.com/sindresorhus/eslint-config-xo/blob/3f8179657736c4c078759416509ca9698e5321e6/esnext.js#L16 too?

main.js
```diff
@@ -21,6 +22,7 @@ const cli = meow(`
 	  --space           Use space indent instead of tabs  [Default: 2]
 	  --no-semicolon    Prevent use of semicolons
 	  --prettier        Conform to Prettier code style
+	  --node-version    Range of Node version to support
```
sindresorhus: `Node` => `Node.js`

readme.md
```diff
@@ -27,6 +27,7 @@ Uses [ESLint](http://eslint.org) underneath, so issues regarding rules should be
 - No need to specify file paths to lint as it lints all JS files except for [commonly ignored paths](#ignores).
 - [Config overrides per files/globs.](#config-overrides)
 - Includes many useful ESLint plugins, like [`unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn), [`import`](https://github.com/benmosher/eslint-plugin-import), [`ava`](https://github.com/avajs/eslint-plugin-ava), [`node`](https://github.com/mysticatea/eslint-plugin-node) and more.
+- Automatically enable or disable rules based on the [engines](https://docs.npmjs.com/files/package.json#engines) of your `package.json`.
```
sindresorhus: > Automatically toggles rules based on the `engines` field in your `package.json`.

How about this?

sindresorhus:
> Do I have a convenient way to determine where are the AVA test files for the current user?

The list you added works if the user just runs AVA with `$ ava`, but they could be running it with `$ ava foo.js` and we wouldn't detect it. Those file patterns might also match users of other test frameworks which doesn't compile like AVA does, so I think this approach is a no-go.

Another approach would be to just run a simple regex for `require('ava')` or `from 'ava';` at the top of the file.

sindresorhus:
> That should be solved once mysticatea/eslint-plugin-node@846e677 get released. We could enable the rule then.

👍

lib/options-manager.js
pvdlg: Id didn't include it becasue it's deprecated. See https://eslint.org/docs/rules/prefer-reflect

pvdlg:
> The list you added works if the user just runs AVA with $ ava, but they could be running it with $ ava foo.js and we wouldn't detect it. Those file patterns might also match users of other test frameworks which doesn't compile like AVA does, so I think this approach is a no-go.
Another approach would be to just run a simple regex for require('ava') or from 'ava'; at the top of the file.

Ok make sense. Should implement this way then?

lib/options-manager.js
sindresorhus: Ah, I didn't realize. We can eventually fix that: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/142

pvdlg:
Actually thinking about the AVA files path issue, I wonder if it's a good to use a regex to detect if AVA is imported.
That would prevent users to toggles rules or to use `overrides` in the case of AVA files.

sindresorhus:
Yeah, let's leave it out for now.

lib/options-manager.js
pvdlg: I think there is a bunch of rules to add from `eslint-plugin-unicorn` and some of them should apply only for specific Node.js version. But that might be better to do that in a different PR.

sindresorhus:
> Like should I revert a78e02e or keep it?

Revert, but put it in a separate branch so we can continue it later. It's not a priority now, but I would like to solve it at some point.

pvdlg:
Ok revert is done

sindresorhus:
This looks excellent. Thanks so much for working on this @pvdlg. I'm really happy about being able to use newer rules without having to manually configure them :)

PR #267, opened 2017-11-17
WIP: Use globby gitignore handling

marionebl:
* closes #236
* ~~failing test cases indicate regression in globby gitignore handling~~ Just faulty usage of globby

## ToDo

* [x] expose `getGitIgnoreFilter` from globby: [globby#56](https://github.com/sindresorhus/globby/pull/56)
* [x] investigate failing test cases
* [x] remove all remnants of `lib/gitignore`

PR #283, opened 2018-01-08
Use globby gitignore handling

### LINKED ISSUE #236: Move the gitignore handling code to `globby`
As an optional feature in `globby`. It was a really difficult thing to implement, so would be nice if everyone could take advantage of and help improve it.

@marionebl @schnittstabil Thoughts?

pvdlg:
Cherry picked the commits from #267, and added minor changes.

Test are passing now.

Fix #267, Fix #236


index.js
```diff
@@ -62,9 +62,9 @@ module.exports.lintText = (str, opts) => {
 	if (opts.filename) {
 		const filename = path.relative(opts.cwd, opts.filename);
 		const isIgnored = multimatch(filename, opts.ignores).length > 0;
-		const isGitIgnored = !optionsManager.getGitIgnoreFilter(opts)(opts.filename);
+		const isGitIgnored = globby.gitignore.sync({cwd: opts.cwd, ignore: opts.ignores});
 
-		if (isIgnored || isGitIgnored) {
+		if (isIgnored || isGitIgnored(opts.filename)) {
```
marionebl: I prefer conditions like this one to be symmetric, so deciding on either

```js
if (fn1(filename) || fn(filename)) {

}
```

**or**

```js
var one = fn1(filename);
var two = fn2(filename);

if (one || two) {

}
```

would help readability here. Doing the execution late could help save some io in some cases but without a proof this has a meaningful impact on performance I'd opt for higher readability (additional binding to intermediary variables).

pvdlg: Actually I made this change from your code because it was really confusing to me why `multimatch(filename, opts.ignores).length > 0` was wrap into a function. It took me several minutes to figurate out the function wrapper was useless.

It might looks more elegant or symmetrical to have `if (isIgnored(opts.filename) || isGitIgnored(opts.filename))` instead of `if (isIgnored || isGitIgnored(opts.filename))` but in terms of cognitive load it's really confusing to wrap something in a function to call it on the next line.

It's always a good idea to improve the formatting or splitting the code for readability, but I'm not convinced that changing the behavior of the code (like wrapping in a function) for readability purpose is so good.

marionebl: We are on the same page then with 

```js
const isIgnored = match(...args1);
const isGitIgnored = match(...args2)(...args3);

if (isIgnored || isGitIgnored) {

}
```

being the best option.

pvdlg: In that situation you would systematically call `globby.gitignore.sync(...args2)(...args3)`. While with the current solution it's called only if `isIgnored` is falsy.

Calling it has a performance cost as it requires to read the `.gitignore` file.
So there is no need to do that if the file is ignored anyway and `isIgnored` is falsy.

I don't think it's a good idea to make the code behave differently and read a file on the disk, in order to make an `if` look symmetrical.

marionebl: Sure, that would be true if for`globby.gitignore.sync(...args2)(...args3)` the second invocation `fn(...args3)` triggered the IO/side effects. 

In the current implemenation over at `globby` [this work](https://github.com/sindresorhus/globby/blob/master/gitignore.js#L84) already is done during the first invocation `fn(...args2)`. 

A performance optimization with impact would be:

```js
if (match(...args) || globby.gitignore.sync(...args2)(...args3)) {

}
```

A bit less terse / maybe more readable

```js
if (match(...args)) {
  return {...};
}

const isGitIgnored = globby.sync(...args2);

if (isGitIgnored(...args3)) {
  return {...};
}
```

pvdlg: I change it for the following:
```js
if (multimatch(filename, opts.ignores).length > 0 ||
  globby.gitignore.sync({cwd: opts.cwd, ignore: opts.ignores})(opts.filename)) {

}
```

But really, code is more readable and easy to understand if it simply does what its meant for. Changing it in a convoluted way for symmetrical reasons just add to the cognitive load of understanding it.

marionebl: We'll have to disagree on this and recognize cognitive load might be caused by different things for each of us. 

One additional thought: an early return might express your **intent** to save work if possible a bit better.

I am fine with the intended change regardless.

sindresorhus:
Awesome. Thanks for working on this @pvdlg and @marionebl 🙌

Issue #205, opened 2017-04-05
Idea: Add `eslint-plugin-security` rules


Rowno:
Just an idea, but some of the [eslint-plugin-security](https://github.com/nodesecurity/eslint-plugin-security) rules might be good candidates to add to XO, like `detect-unsafe-regex`, `detect-eval-with-expression` and `detect-non-literal-require`.

sindresorhus:
I looked at it awhile ago and the only interesting rule then was `detect-unsafe-regex`. The other ones are just too naive with too many potential false positives. Maybe also `detect-eval-with-expression`, but everyone knows `eval` is bad, so it won't really help much. I used to have [`no-dynamic-require`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-dynamic-require.md) enabled which is similar to `detect-non-literal-require`, but I ended removing it (https://github.com/sindresorhus/xo/commit/29a5263742f31109cd3fdfd19eb4b3ea0d27de4b), as it was just too annoying. The only time I used dynamic require was when it was the only choice.

I'll keep this open for further discussion, but I'm not entirely convinced yet about adding this plugin just for one rule.

Rowno:
It would be awesome just to have `detect-unsafe-regex` added to XO (in whatever form) since it's such a common security issue in Node.js. I can port it to `eslint-plugin-unicorn` if you want it.

sindresorhus:
> It would be awesome just to have detect-unsafe-regex added to XO (in whatever form) since it's such a common security issue in Node.js.

Agreed, but there are some problems with it, see: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/153

PR #285, opened 2018-01-19
Add detect-unsafe-regex rule

### LINKED ISSUE #227: Add regex plugins
## [eslint-plugin-no-unsafe-regex](https://github.com/kgryte/eslint-plugin-no-unsafe-regex)
"unsafe regex" as defined by [substack/safe-regex](https://github.com/substack/safe-regex)
> detect possibly catastrophic, exponential-time regular expressions

[Examples](https://github.com/substack/safe-regex/blob/93d23dd464f45f0ea68342b9cba8da2fdf97c7f2/test/regex.js#L22-L28) of bad regexes:

```js
/^(a?){25}(a){25}$/,
RegExp(Array(27).join('a?') + Array(27).join('a')),
/(x+x+)+y/,
/foo|(x+x+)+y/,
/(a+){10}y/,
/(a+){2}y/,
/(.*){1,32000}[bc]/
```

## [eslint-plugin-optimize-regex](https://github.com/BrainMaestro/eslint-plugin-optimize-regex)

[Example](https://github.com/BrainMaestro/eslint-plugin-optimize-regex/blob/c84fcc9/tests/lib/rules/optimize-regex.js#L28)

```
/[a-zA-Z_0-9][A-Z_\\da-z]*\\e{1,}/ can be optimized to /\\w+e+/
```

bdougherty:
Decided to go with just a warning for the unoptimized regex plugin since that is the example on the repo. I'll change it if you think it should be an error instead.

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -16,7 +16,8 @@ module.exports = {
 		'unicorn',
 		'promise',
 		'import',
-		'node'
+		'node',
+		'security'
 	],
 	extends: [
 		'plugin:ava/recommended',
@@ -81,6 +82,7 @@ module.exports = {
 		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin". See https://github.com/mysticatea/eslint-plugin-node/issues/96
 		// 'node/shebang': 'error',
 		'node/no-deprecated-api': 'error',
-		'node/exports-style': ['error', 'module.exports']
+		'node/exports-style': ['error', 'module.exports'],
+		'security/detect-unsafe-regex': 'error'
 	}
 };
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
 		"eslint-plugin-node": "^5.2.1",
 		"eslint-plugin-prettier": "^2.3.1",
 		"eslint-plugin-promise": "^3.6.0",
+		"eslint-plugin-security": "^1.4.0",
 		"eslint-plugin-unicorn": "^3.0.1",
 		"get-stdin": "^5.0.0",
 		"globby": "^7.1.1",


bdougherty:
Hmm didn't realize the optimize regex plugin requires node >= 6. Let me know if you are ok with bumping the xo requirement or if I should remove that plugin.

pvdlg:
Maybe we can use [eslint-plugin-security#detect-unsafe-regex](https://github.com/nodesecurity/eslint-plugin-security#detect-unsafe-regex) instead of `no-unsafe-regex/no-unsafe-regex` as the `eslint-plugin-security` seems better maintained and contains other rules we might want to enable.

Regarding the optimize regex, I'm not sure 1 rule would justify to drop node 4 support (at least while node 4 is still in [LTS support](https://github.com/nodejs/Release#release-schedule)).

sindresorhus:
We can not drop Node.js 4 support yet.

bdougherty:
Updated to use [eslint-plugin-security#detect-unsafe-regex](https://github.com/nodesecurity/eslint-plugin-security#detect-unsafe-regex).

Found another discussion about that plugin in #205 where you weren't sure about adding the full plugin for just the one rule. Happy to add an unsafe regex rule to eslint-plugin-unicorn if that would be preferable.

sindresorhus:
> Happy to add an unsafe regex rule to eslint-plugin-unicorn if that would be preferable.

Yeah, that would be preferable, seeing how it's only a few lines of glue code to the `safe-regex` module, and that the `eslint-plugin-security` plugin doesn't seem very actively maintained. I think it would also be a good rule to use Selectors: https://eslint.org/docs/developer-guide/selectors You can just copy-paste this rule and do your changes: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/regex-shorthand.js :)

bdougherty:
sindresorhus/eslint-plugin-unicorn#146 created.

PR #286, opened 2018-01-20
Add JSDoc plugin

### LINKED ISSUE #282: Add --jsdoc CLI option to enable jsdoc linting
## Proposal

I think that it would be a great addition to let users enable this essential ESLint feature easily, without worrying about adding huge rules inside the `package.json`.

## Example usage

Just pass the flag
```bash
xo --jsdoc
```
or enable it via `package.json`
```js
"xo": {
  "jsdoc": true
}
```

## Rules added

ESlint additional config when the option is provided
```js
"xo": {
  "space": true,
  "plugins": ["jsdoc"],
  "rules": {
    "jsdoc/newline-after-description": "warn",
    "jsdoc/require-description-complete-sentence": "warn",
    "jsdoc/require-hyphen-before-param-description": "warn",
    "require-jsdoc": [
      "warn", {
        "require": {
          "FunctionDeclaration": true,
          "MethodDefinition": true,
          "ClassDeclaration": true,
          "ArrowFunctionExpression": true,
          "FunctionExpression": true // Currently throws an exception see https://github.com/eslint/eslint/issues/9825
        }
      }
    ],
    "valid-jsdoc": ["warn", {
      "requireReturn": false,
      "prefer": {
        "arg": "param",
        "argument": "param",
        "arguments": "extends",
        "const": "constant",
        "defaultvalue": "default",
        "desc": "description",
        "exception": "throws",
        "file": "overview",
        "fileoverview": "overview",
        "fires": "emits",
        "func": "function",
        "prop": "property",
        "return": "returns",
        "virtual": "abstract",
        "yeld": "yelds"
      },
      "preferType": {
        "array": "Array",
        "Boolean": "boolean",
        "date": "Date",
        "Number": "number",
        "object": "Object",
        "regex": "RegExp",
        "String": "string"
      }
    }]
  }
}
```
Let's discuss about the defaults, and then in case I will open a PR.

## JSDoc related eslint plugins and rules:
[eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) by @gajus
[require-jsdoc](https://eslint.org/docs/rules/require-jsdoc) by @eslint
[valid-jsdoc](https://eslint.org/docs/rules/valid-jsdoc) by @eslint

pvdlg:
Related sindresorhus/eslint-config-xo#45

Fix #282

I added the plugin and rules suggested in #282. I didn't add the `--jsdoc` option for the reasons explained in [this comment](https://github.com/sindresorhus/xo/issues/282#issuecomment-356035657):

> As the `valid-jsdoc` rule is used when a JSDoc bloc is defined, in the end the `jsdoc` flag would just enable the `require-jsdoc` rule forcing to write JSDoc for every declaration in the code.

> Enforcing 100% JSDoc is not really something very common from what I see in the open source community (AFAIK ESLint is the only major project with near 100% JSDoc coverage).

> In addition, it would be difficult to find the best default config for the rule. Some might want a warning, some an error. Some might want a JSDoc block for function expression and declaration, some for declaration only etc...

> So I'm not sure adding a flag is justified for a case that is not really common, and requires additional config anyway. It's probably better to just set the rule in XO:
```json
{
  "xo": {
    "rules": {"require-jsdoc": ["warn", {...}]}
  }
}
```

@sindresorhus if you think the `--jsdoc` flag makes sense I can add it. Not sure what the `require-jsdoc` options would be though.

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -16,7 +16,8 @@ module.exports = {
 		'unicorn',
 		'promise',
 		'import',
-		'node'
+		'node',
+		'jsdoc'
 	],
 	extends: [
 		'plugin:ava/recommended',
@@ -80,6 +81,9 @@ module.exports = {
 		// 'node/shebang': 'error',
 		'node/no-deprecated-api': 'error',
 		'node/exports-style': ['error', 'module.exports'],
+		'jsdoc/newline-after-description': 'warn',
+		'jsdoc/require-description-complete-sentence': 'warn',
+		'jsdoc/require-hyphen-before-param-description': 'warn',
 		// Disabled by default (overrides `plugin:unicorn/recommended`), will be enabled if supported by the Node.js version
 		'unicorn/prefer-spread': 'off',
 		'unicorn/no-new-buffer': 'off'
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
 		"eslint-formatter-pretty": "^1.0.0",
 		"eslint-plugin-ava": "^4.2.0",
 		"eslint-plugin-import": "^2.0.0",
+		"eslint-plugin-jsdoc": "^3.3.1",
 		"eslint-plugin-no-use-extend-native": "^0.3.2",
 		"eslint-plugin-node": "^5.2.1",
 		"eslint-plugin-prettier": "^2.3.1",


sindresorhus:
> if you think the --jsdoc flag makes sense I can add it. Not sure what the require-jsdoc options would be though.

No, I agree with you. We should not add it.

pvdlg:
Just rebased. As we won't add the `--jsdoc` flag it should be good to go

PR #291, opened 2018-01-29
Revert JSDoc plugin

pvdlg:
Per [comment](https://github.com/sindresorhus/xo/issues/282#issuecomment-359282802) those new rules have several bugs.

```js
/**
 * Sentence with a word stating with a
 * Capital letter.
 */
```
Returns an error, even though it should be valid to break the line and sometimes the break happen before a capitalize word.

```js
/**
 * List of things:
 * - elem 1
 * - elem 2
 */
```
Returns an error even though doing a list should be perfectly valid in JSDoc.

```js
/**
 * Code example:
 * ```js
 * console.log(test)
 * ```
 */
```

Once again returns an error even though perfectly valid.

All the following situation might create an issue:
- Using capital letter in the middle of a sentence
- Using list
- Using colon (`:`) + new line to describe something
- Using code example
- Using mardown

We should remove those rules until they work properly.
@simonepri as you requested those rules in the first place, maybe you could open issues with the `eslint-plugin-jsdoc` project?

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -16,8 +16,7 @@ module.exports = {
 		'unicorn',
 		'promise',
 		'import',
-		'node',
-		'jsdoc'
+		'node'
 	],
 	extends: [
 		'plugin:ava/recommended',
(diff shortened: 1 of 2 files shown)


simonepri:
Hi @pvdlg
Sorry for the troubles.

I don't think that this:
```
/**
 * Sentence with a word stating with a
 * Capital letter.
 */
```
can be fixed easily.

Because any *easy* fix would not report this :
```
/**
 * A sentence that is not ended
 * New sentence.
 */
```

pvdlg:
Let's open issues with `eslint-plugin-jsdoc`, see what get fixed and re-evaluate later on what rule is stable enough to be used.

PR #295, opened 2018-02-13
Update globby to the latest version

pvdlg:


pvdlg:
@sindresorhus I'll let you do the merge, as it seems Travis is not configured yet to work with the new orga.

PR #297, opened 2018-02-13
Disable `enforceForRenamedProperties` for `prefer-destructuring`

pvdlg:
Disabling due to eslint/eslint#9970


sindresorhus:
I know ESLint that it's expected behavior, but I find the behavior annoying, so I'm gonna merge this regardless.

pvdlg:
Yep, agreed, this is why I kept the PR open after eslint response

PR #300, opened 2018-02-14
Do not add indent rule when `prettier` option is true

pvdlg:
In the specific case of both `prettier` and `space` being `true`, the indent rules is customized by XO.

Rules defined in the `rules` property override the one defined in a config file (in this case `eslint-config-prettier`). So it ends up in a situation in which the number of space to use is handled by both `prettier` and the eslint `indent` rules. In some cases it creates a conflict.

With this change the `indent` rule is set only when `prettier` is `false`.

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -182,7 +182,7 @@ const buildConfig = opts => {
 		}
 	}
 
-	if (opts.space) {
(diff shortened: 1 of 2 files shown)


pvdlg:
It's the PR number 300 and the 100th to be closed. I don't why I notice these weird coincidences :)

Issue #309, opened 2018-03-16
enhancement: support option --print-config and --debug


aladdin-add:

sindresorhus:
Can you elaborate on exactly what these options would do?

Aladdin-ADD:
#### `--debug`

This option outputs debugging information to the console. This information is useful when you're seeing a problem and having a hard time pinpointing it. The ESLint team may ask for this debugging information to help solve bugs.

#### `--print-config`

This option outputs the configuration to be used for the file passed. When present, no linting is performed and only config-related options are valid.

sindresorhus:
I think both of these would be covered by https://github.com/xojs/xo/issues/188.

fregante:
Closing in favor of https://github.com/xojs/xo/issues/188

Issue #272, opened 2017-11-27
Add a CLI option for exporting a vanilla ESLint config


jamestalmage:
See https://github.com/jamestalmage/xo-with-webstorm

Lots of editors support ESLint directly. Not that many support XO. We should add a `xo --export-config` command that outputs a `.eslintrc` file, and ensures all the correct packages are installed at the top level.

```
$ xo --export-config
This will overwrite the existing .eslintrc file. Ok? (y/n) y

Some of XO's modules need to be installed as top level dependencies for this to work.
Enter the following command:

   npm install --save-dev eslint-config-xo eslint-plugin... 
```

sindresorhus:
Related to #122.

I'm ok with adding this flag.

pvdlg:
XO has it own implementation of `overrides` which mean when we export the ESLint config we would have to translate from XO implementation to ESLint.

As it is now [supported by ESLint](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns), shouldn't we remove the XO `override` implementation and use ESLint one instead?

That would make this feature slightly easier to implement

slikts:
Generating an ESLint config that would extend the XO config would make editor integrations work out-of-the-box for all the editors that support it; instead, now XO-specific [editor extensions][0] are required. I understand this is a legacy issue, but it's a significant letdown.

[0]: https://github.com/SamVerschueren/vscode-linter-xo

fregante:
The new Flat config format essentially covers these needs:

- If you want an all-in-one dependency, install and run `xo`
- If you want an eslint config, install `eslint-config-xo`, which will _soon_ include all the necessary dependencies

Since this issue is now 5 years old with little progress/interest + Prettier and TypeScript having become dependencies (further complicating the endeavor), I'm closing in favor of 

- https://github.com/xojs/xo/issues/702

PR #312, opened 2018-03-23
Add `print-config` option

teehemkay:
Essentially mimics `eslint` [`--print-config`](https://eslint.org/docs/user-guide/command-line-interface#--print-config) option.

Related to #272 and #309.

As an added bonus, this feature enables the use of `xo` as a drop-in replacement for the global `eslint` executable when using the [`eslint-javascript`](http://www.flycheck.org/en/latest/languages.html#javascript) linter of FlyCheck, the popular GNU Emacs syntax checking tool.


pvdlg:
Thanks for the PR!

XO as as ESLint allows to create specific config for a subset of files with `overrides`. However they work quite differently:
- Eslint `overrides` cannot contains `extends` XO's can
- XO `overrides` can contain XO options (`space`, `semicolon`, etc...) but ESLint can't

How do you handle those cases? Is your PR produces a valid ESLint config for the following XO config?
```json
{
  "overrides": [
    {
      "files": "test/*.js",
      "extends": "some-custom-config",
      "esnext": false,
      "space": 3
    },
    {
      "files": "test/foo.js",
      "esnext": true
    }
  ]
}
```

teehemkay:
I was also thinking about actually having two options: 
The first one would return the config as pre-processed by XO (_ie_, before sending it to Eslint). This would enable debugging the XO pre-processing. Suggested name would be something like `--print-xo-config` 
The second one would be the current PR.
Any thoughts about this suggestion?

pvdlg:
In order to debug the XO pre-processing we can use [debug](https://github.com/visionmedia/debug) with possible a `--debug` CLI argument. But that would be another PR.

So I don't think there is really a need for 2 options. The objective is to generate a single config file that can be used directly with `ESLint`.

I try to look at this functionality a while ago and I couldn't figure out a way to "translate" XO certain `overrides` use case into ESLint `overrides` as they have some differences.

teehemkay:
Sorry for the confusion.

As I wrote in my initial comment, this PR

> Essentially mimics `eslint` [`--print-config`](https://eslint.org/docs/user-guide/command-line-interface#--print-config) option.

It's only _related_ to #272 and #309 

The PR pre-processes `xo` options and then pass them to the `eslint` function that implements the `print-config` option. It's the later that does the heavy lifting of generating the config.

Since `print-config` is only concerned with a single file, we only need to make sure to pull in any _overriding_ option _that needs to be applied_ to the file. 

So, unless I'm mistaken, the `eslint` override feature does not seem to be relevant here.

My motivation in implementing this PR, is to enable `xo` to act as as a drop-in replacement (sill some work to do) for `eslint` when using ['FlyCheck`](http://www.flycheck.org/en/latest/languages.html#javascript) which seems to be the most popular  `Emacs` linter.

teehemkay:
Closed the PR since it doesn't seem to be going anywhere.

PR #326, opened 2018-05-17
Add support for TypeScript and document usage with Flow

### LINKED ISSUE #315: Improve documentation for usage with Flow and TypeScript
Although XO does not officially support Flow or TypeScript, they work pretty well with some configuration:

**Flow:**

`npm install --save-dev babel-eslint`
```json
"xo": {
	"parser": "babel-eslint"
}
```

**TypeScript:**

To get TypeScript to work, the `.ts` extension needs to be manually added [here](https://github.com/xojs/xo/blob/master/lib/options-manager.js#L28) first.

`npm install --save-dev typescript-eslint-parser`
```json
"xo": {
	"parser": "typescript-eslint-parser",
	"parserOptions": {
		"typescript": true
	}
}
```

Supporting the `.ts` extension natively would of course be much better than editing the `xo` package in your `node_modules`, and it would be nice if this was all documented in the `readme`.

jorgegonzalez:
Fixes #315

Allows TypeScript and Flow files to be parsed by XO by setting `"flow": true` or `"typescript": true` in the config.

I think the error messages can be improved a bit. And I'm not 100% sure if that's where this particular block of code should go, but it works.

Also there is probably a better way to get the `devDependencies` list.

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -26,19 +26,6 @@ module.exports = {
 			'electron',
 			'atom'
 		],
-		'import/resolver': {
-			'node': {
-				'extensions': [
-					'.js',
-					'.jsx',
-					'.ts',
-					'.tsx'
-				]
-			}
-		},
-		'import/parsers': {
-			'typescript-eslint-parser': ['.ts', '.tsx']
-		}
 	},
 	rules: {
 		'no-use-extend-native/no-use-extend-native': 'error',
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -26,6 +26,16 @@ module.exports = {
 			'electron',
 			'atom'
 		],
+		'import/resolver': {
+			'node': {
+				'extensions': [
+					'.js',
+					'.jsx',
+					'.ts',
+					'.tsx'
+				]
+			}
+		},
 		'import/parsers': {
 			'typescript-eslint-parser': ['.ts', '.tsx']
 		}
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -25,7 +25,7 @@ module.exports = {
 		'import/core-modules': [
 			'electron',
 			'atom'
-		],
+		]
 	},
 	rules: {
 		'no-use-extend-native/no-use-extend-native': 'error',
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -25,7 +25,17 @@ module.exports = {
 		'import/core-modules': [
 			'electron',
 			'atom'
-		]
+		],
+		'import/resolver': {
+			'node': {
+				'extensions': [
+					'.js',
+					'.jsx',
+					'.ts',
+					'.tsx'
+				]
+			}
+		}
 	},
 	rules: {
 		'no-use-extend-native/no-use-extend-native': 'error',
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -82,21 +82,8 @@ module.exports = {
 		'node/process-exit-as-throw': 'error',
 		// Disabled as the rule doesn't exclude scripts executed with `node` but not referenced in "bin". See https://github.com/mysticatea/eslint-plugin-node/issues/96
 		// 'node/shebang': 'error',
-		'node/no-deprecated-api': 'error',
+		'node/no-deprecated-api': 'error'
 		// Disabled because it causes too much churn and will be moot when we switch to ES2015 modules
 		// 'node/exports-style': ['error', 'module.exports']
-		'typescript/adjacent-overload-signatures': 'error',
-		'typescript/class-name-casing': 'error',
-		'typescript/member-delimiter-style': 'error',
-		'typescript/no-array-constructor': 'error',
-		'typescript/no-empty-interface': 'error',
-		'typescript/no-non-null-assertion': 'error',
-		'typescript/no-triple-slash-reference': 'error',
-		'typescript/no-type-alias': 'error',
-		'typescript/no-unused-vars': 'error',
-		'typescript/no-use-before-define': 'error',
-		'typescript/no-var-requires': 'error',
-		'typescript/prefer-namespace-keyword': 'error',
-		'typescript/type-annotation-spacing': 'error'
 	}
 };
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -10,13 +10,12 @@ module.exports = {
 	},
 	// -- end repeat
 	plugins: [
-		'ava',
-		'import',
 		'no-use-extend-native',
-		'node',
-		'promise',
-		'typescript',
+		'ava',
 		'unicorn',
+		'promise',
+		'import',
+		'node'
 	],
 	extends: [
 		'plugin:ava/recommended',
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -26,16 +26,6 @@ module.exports = {
 			'electron',
 			'atom'
 		],
-		'import/resolver': {
-			'node': {
-				'extensions': [
-					'.js',
-					'.jsx',
-					'.ts',
-					'.tsx'
-				]
-			}
-		},
 		'import/parsers': {
 			'typescript-eslint-parser': ['.ts', '.tsx']
 		}
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -10,12 +10,13 @@ module.exports = {
 	},
 	// -- end repeat
 	plugins: [
-		'no-use-extend-native',
 		'ava',
-		'unicorn',
-		'promise',
 		'import',
-		'node'
+		'no-use-extend-native',
+		'node',
+		'promise',
+		'typescript',
+		'unicorn',
 	],
 	extends: [
 		'plugin:ava/recommended',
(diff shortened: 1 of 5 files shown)


pvdlg:
Instead of looking into the dev dependencies could we do something like that?
- Add `ts` and `tsx` to the default extension list we parse
- For files with a `.ts` or `.tsx` extension automatically set the parser to `typescript-eslint-parser`
- Add a paragraph to the documentation explaining that in order to use flow, the parser must must be set to `babel-parser` (as we don't have a way to detect if a file is written in Flow or regular JS and set the parser automatically)

That would avoid adding the `typescript` and `flow` config options which are not really useful. That would also make things works out of the box for users (at least in the case of typescript).

jorgegonzalez:
I agree.

jorgegonzalez:
XO will show these errors if the proper dependencies are not installed:

image

image

image

index.js
```diff
@@ -102,6 +102,10 @@ module.exports.lintFiles = (patterns, opts) => {
 			});
 		}
 
+		if (paths.filter(path => path.includes('.ts')).length > 0) {
```
pvdlg: You should use [path.extname](https://nodejs.org/api/path.html#path_path_extname_path) here instead of `includes`. I guess it should also includes `.tsx` files as well.

pvdlg:
Ultimately we might consider using TSLint rather (or in combination of) to lint `.ts` and `.tsx` files but that would be another issue/PR/discussion.

jorgegonzalez:
There is a benefit in applying ESLint rules to TypeScript, as TSLint is missing many of the nicer ESLint plugins (like `eslint-plugin-unicorn`), and it can be a little wonky in my experience.

pvdlg:
Indeed. That would probably be a combination of both.

jorgegonzalez:
Hmmm getting an `(import/no-unresolved)` error for `.ts` files.

readme.md
```diff
@@ -266,6 +266,35 @@ Enforce ES2015+ rules. Disabling this will make it not *enforce* ES2015+ syntax
...
 *ES2015+ is parsed even without this option. You can already use ES2017 features like [`async`/`await`](https://github.com/lukehoban/ecmascript-asyncawait).
 
+## Flow and TypeScript
+
+### Flow
+
+To enable parsing of Flow-typed files, install `babel-eslint` and add it to the config:
+
+```
+npm i -D babel-eslint
```
sindresorhus: Use long-hand for documentation: `npm install --save-dev babel-eslint`

pvdlg:
> Hmmm getting an import/no-unresolved error for .ts files. But only if the extension is excluded in the import

Sounds like something to report to `eslint-plugin-import` project.

sindresorhus:
I think you need to set the [`import/parser` option](https://github.com/benmosher/eslint-plugin-import#importparsers) too.

readme.md
```diff
@@ -266,6 +266,35 @@ Enforce ES2015+ rules. Disabling this will make it not *enforce* ES2015+ syntax
...
+
+```json
+{
+	"xo": {
+		"parser": "babel-eslint"
+	}
+}
+```
+
+### TypeScript
```
sindresorhus: Place TypeScript first since it's more popular.

jorgegonzalez:
Oh cool, setting `import/parsers` seems like a better solution than what I had done previously.

config/plugins.js
```diff
@@ -25,7 +25,17 @@ module.exports = {
 		'import/core-modules': [
 			'electron',
 			'atom'
-		]
+		],
+		'import/resolver': {
```
pvdlg: Shouldn't it be as follow?
```
import/parsers: {
    typescript-eslint-parser: [ .ts, .tsx ]
}
```

Also should we add that only for `ts`/`tsx` files similarly to `parser = 'typescript-eslint-parser'`?

jorgegonzalez: Yep.

jorgegonzalez:
Ah you're right, sorry. Was testing with the wrong files

config/plugins.js
```diff
@@ -25,7 +25,10 @@ module.exports = {
 		'import/core-modules': [
 			'electron',
 			'atom'
-		]
+		],
+		'import/parsers': {
```
pvdlg: I think that should be set dynamically, only when linting a `ts` or `tsx` file. Otherwise `typescript-eslint-parser` might become needed even the current file is a regular JS file.
So in `index.js`:
```js
		if (paths.filter(filePath => isTSFile(path.extname(filePath))).length > 0) {
			opts.parser = 'typescript-eslint-parser';
            opts['import/parsers']['typescript-eslint-parser'] = ['.ts', '.tsx'];
		}
```

config/plugins.js
jorgegonzalez: Actually still need this or the import/unresolved rule throws.

pvdlg: So it should be done dynamically in the `index.js` only for `ts`/`tsx` files I imagine

jorgegonzalez: I don't think it's an issue unless someone is intentionally mixing `.js` and `.ts` files in a directory.

jorgegonzalez:
Should we explore support for [`eslint-plugin-typescript`](https://github.com/nzakas/eslint-plugin-typescript) as well?

config/plugins.js
pvdlg: Could you move the `import/parsers` and `import/resolver` in `index.js` so it will be enabled only for `ts` and `tsx` files? This way that would avoid creating additional load and add a dependency to `typescript-eslint-parser` for non TS file.
```js
if (paths.filter(filePath => isTSFile(path.extname(filePath))).length > 0) {
  opts.parser = 'typescript-eslint-parser';
  opts['import/resolver'].node.extensions = ['.js', '.jsx', '.ts', 'tsx'];
  opts['import/parsers'].['typescript-eslint-parser'] = ['.ts', 'tsx'];
}
```

sindresorhus:
> Should we explore support for eslint-plugin-typescript as well?

👍

sindresorhus:
> The other option is to call Eslint and TSLint API from XO and merge the result object.

I'm not sure we want that. TSLint is not that actively maintained and it's very buggy (talking from experience using it in TS projects). I'm holding out on ESLint being the winner for TS projects too.

config/plugins.js
jorgegonzalez: Okey dokey

index.js
```diff
@@ -102,6 +102,20 @@ module.exports.lintFiles = (patterns, opts) => {
 			});
 		}
 
+		const isTSFile = ext => ext === '.ts' || ext === '.tsx';
+
+		if (paths.filter(filePath => isTSFile(path.extname(filePath))).length > 0) {
+			opts.parser = 'typescript-eslint-parser';
+			opts['settings'] = {
```
pvdlg: I don't think this is correct as with the current code the entire `settings` value will be overridden.
We want to add/merge only the last leaf (`extensions` and `typescript-eslint-parser`).
So if the user config already contains something like:
```js
{
settings: {
  'import/resolver': {
    node: {extensions: ['.js', '.jsx', '.xyz']}
  }
}
}
```
We want to end up with:
```js
{
settings: {
  'import/resolver': {
    node: {extensions: ['.js', '.jsx', '.xyz', '.ts', '.tsx']}
  }
}
}
```

pvdlg: `opts['settings']` should be `opts.settings` for consistency.

jorgegonzalez: You right

index.js
```diff
@@ -102,6 +102,20 @@ module.exports.lintFiles = (patterns, opts) => {
 			});
 		}
 
+		const isTSFile = ext => ext === '.ts' || ext === '.tsx';
+
+		if (paths.filter(filePath => isTSFile(path.extname(filePath))).length > 0) {
```
pvdlg: Instead of `filter` + `.length > 0` you'll get better performances with `find` as it will stop on the first item found:
```js
if (paths.find(filePath => isTSFile(path.extname(filePath))))
```

index.js
```diff
@@ -102,6 +102,33 @@ module.exports.lintFiles = (patterns, opts) => {
...
+		if (paths.find(filePath => isTSFile(path.extname(filePath)))) {
+			opts.parser = 'typescript-eslint-parser';
+
+			const extensions =
+				opts.settings &&
+				opts.settings['import/resolver'] &&
+				opts.settings['import/resolver'].node
+					? opts.settings['import/resolver'].node.extensions
+					: [];
+
```
jorgegonzalez: There is probably a cleaner way to do this.

index.js
```diff
@@ -102,6 +102,34 @@ module.exports.lintFiles = (patterns, opts) => {
 			});
 		}
 
+		const isTSFile = ext => ext === '.ts' || ext === '.tsx';
+
+		if (paths.find(filePath => isTSFile(path.extname(filePath)))) {
+			opts.parser = 'typescript-eslint-parser';
+
+			const extensions =
```
jorgegonzalez: There is probably a better way to do this

pvdlg: You should use `mergeWith` like in the rest of the codebase: https://github.com/xojs/xo/blob/master/lib/options-manager.js#L168

config/plugins.js
```diff
@@ -10,12 +10,13 @@ module.exports = {
...
-		'no-use-extend-native',
 		'ava',
-		'unicorn',
-		'promise',
 		'import',
-		'node'
+		'no-use-extend-native',
+		'node',
+		'promise',
+		'typescript',
```
pvdlg: This is different than the scope of this PR. If we start using `eslint-plugin-typescript` this should be done in a different PR, that might potentially be released later when the plugin is stable.

jorgegonzalez: I disagree. It's pretty usable right now, just missing a lot of rules.

jorgegonzalez: "Early stages" does not imply "unstable" or "not production ready". Recent [npm stats](https://npm-stat.com/charts.html?package=eslint-plugin-typescript&package=xo) indicate that users have confidence in its current usability.

jorgegonzalez:
@pvdlg I really appreciate your help and all the code reviews!

sindresorhus:
@jorgegonzalez I've made https://github.com/xojs/eslint-config-xo-typescript now. Can you mention it in the readme?

jorgegonzalez:
@sindresorhus Do you also want me to mention [xojs/eslint-config-xo-flow](https://github.com/xojs/eslint-config-xo-flow)?

jorgegonzalez:
Does anything else need to be done to merge this?

pvdlg:
It seems we decided to go with the option of using `eslint-config-xo-typescript` and `eslint-config-xo-flow`. So the code changes in `index.js` should be removed.
The only code change to keep is to add `.ts` and `.tsx` to `DEFAULT_EXTENSION`.

The doc should also be modified to simply:
- provide a link to [eslint-config-xo-flow#use-with-xo](https://github.com/xojs/eslint-config-xo-flow#use-with-xo) for Flow
- provide a link to [eslint-config-xo-typescript#use-with-xo](https://github.com/xojs/eslint-config-xo-typescript#use-with-xo) for Typescript

PR #343, opened 2018-07-23
Fix links in documentation

noahbrenner:
Just a little polishing:
* One page was moved to a new URL.
* A fragment link in the README needed to be updated.
* Most of the URLs can use `https` instead of `http`
* Links in the contributor covenant didn't match the source and one was unnecessarily specified twice (as the link text and as the href).

If you'd like any adjustments or want these commits squashed, just let me know!


sindresorhus:
Good cleanup! Thanks :)

PR #345, opened 2018-07-24
Respect `nodeVersion` option set in `override` block

pvdlg:
With the following configuration:
```json
{
	"xo": {
		"nodeVersion": ">=6.0.0",
		"overrides": [
			{
				"files": "**/transpile/**",
				"nodeVersion": ">=8.0.0"
			}
		]
	}
}
```

The files in `**/transpile/**` were wrongly analyzed with `nodeVersion: '>=6.0.0'` because the `nodeVersion` in `overrides` was ignored. The code was mistakenly expecting the option to be set with `engines: {node: '>=8.0.0'}`.

With this PR the config above works as espected.


sindresorhus:
Looks good :)

PR #351, opened 2018-09-10
Move default ESLint cache to node_modules/.cache/xo.

### LINKED ISSUE #335: caching breaks rule `import/named`.
Create `test1.js`:
```js
export const value = 'test1';
```

Create `test2.js`:
```js
import {value} from './test1';

console.log(value);
```
Run `xo`, it correctly does not report any errors.  Now rename the export in `test1.js`:
```
export const value2 = 'test1';
```

Run `xo` again, it fails to report any errors.  Run `touch test2.js && xo`, you will see the correct error:
```

  test2.js:1:9
  ✖  1:9  value not found in ./test1  import/named

  1 error
```

Locally I just cleared the ~/.xo-cache folder so this resolved the issue for me.  Secondary issue, have you considered using `./node_modules/.cache/xo`?  That would have made it easier for me to find/clear the cache.

coreyfarrell:
Use find-cache-dir to follow same standard used by ava and nyc.  This
makes it easier for end-user to find/clear the ESLint cache in case of
caching validation error.

Fixes #335


lib/options-manager.js
```diff
@@ -29,7 +29,7 @@ const DEFAULT_EXTENSION = [
 const DEFAULT_CONFIG = {
 	useEslintrc: false,
 	cache: true,
-	cacheLocation: path.join(os.homedir() || os.tmpdir(), '.xo-cache/'),
+	cacheLocation: findCacheDir({name: 'xo'}),
```
coreyfarrell: Can `xo` be run against a stand-alone script (IE not part of any module)?  If `find-cache-dir` fails to find any package.json then it will return `null`.  If needed I can re-add the original cacheLocation value as a fallback, but no point if `xo` cannot be run in that manner.

sindresorhus: > Can xo be run against a stand-alone script (IE not part of any module)?

Yes

coreyfarrell: Fixed

lib/options-manager.js
```diff
@@ -1,11 +1,11 @@
 'use strict';
-const os = require('os');
```
pvdlg: Shouldn't this import statement stays as `os` is used on line 32?

coreyfarrell: Yes, sorry I missed this, apparently I forgot to rerun 'npm test' after I re-added the directory used when xo is run from outside a project.

Also: I'm just noticing it doesn't appear this repo runs CI on PR's.

coreyfarrell:
Sorry for the delay I just realized that I failed to 'git push' my fix which re-added the import of `os`.

PR #356, opened 2018-12-18
Allow xo configuration in extends

### LINKED ISSUE #352: xo shareable configs?
Should we allow shareable configs to use top level rules such as "prettier" and "space" in shareable XO configs?  It would set the option for xo but delete it before passing along to eslint?

Otherwise I have to do this in all my projects:

```
"xo": {
  "prettier": true,
  "space": true,
  "extends": [ "xo-lass" ]
}
```

tkesgar:
Allows overriding xo-specific (not ESLint) configurations in extends.

Fixes #352

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -199,6 +199,59 @@ const buildConfig = options => {
 	);
 	const spaces = normalizeSpaces(options);
 
+	// Copy options, since it will be modified
+	options = Object.assign({}, options);
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for working on this. This will need some tests.

tkesgar:
Sorry I haven't done any work on this. Looks like the changes needs to be readjusted as well after two years.

PR #359, opened 2018-12-22
Add `eslint-plugin-eslint-comments` ESLint plugin

sindresorhus:
https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/

FYI, @mysticatea :)

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -16,7 +16,8 @@ module.exports = {
 		'unicorn',
 		'promise',
 		'import',
-		'node'
+		'node',
+		'eslint-comments'
 	],
 	extends: [
 		'plugin:ava/recommended',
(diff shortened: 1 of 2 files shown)


sindresorhus:
@mysticatea Why does `no-unused-disable` exist when ESLint CLIEngine has the `reportUnusedDisableDirectives` option? Have you tried this option?

mysticatea:
`reportUnusedDisableDirectives` option exists, it's the reason `eslint-comments/recommended` preset does not include `eslint-comments/no-unused-disable` rule.

On the other hand, `eslint-comments/no-unused-disable` rule still exists because we cannot specify `reportUnusedDisableDirectives` option with `eslintrc` config.

mysticatea:
As a side note, `eslint-comments/no-unused-disable` is using `reportUnusedDisableDirectives` option internally.

PR #360, opened 2018-12-31
CI updates

brandon93s:
- Migrate Windows build to Travis CI
- Cross-platform test fix

Successful build: https://travis-ci.org/brandon93s/xo/builds/473886929


.travis.yml
```diff
@@ -1,9 +1,10 @@
+os:
+  - linux
+  - windows
 language: node_js
 node_js:
   - '10'
   - '8'
   - '6'
-before_install:
-  - npm install --global npm
```
brandon93s: Best guess is this was put in place to get around issues with early versions of npm: https://github.com/xojs/xo/commit/c9b9b8431e6d0383d7c42a2e9d9ac3aff08a45c6. Removing, as it should no longer be necessary and provides better test coverage to use the version that ships with each node version.

sindresorhus: Correct

PR #365, opened 2019-01-13
Verbose flag implementation

### LINKED ISSUE #188: Verbose mode
Would be useful to have an `--verbose` mode that outputs a lot of debug information:

- Which files are ignored by our default ignore filter
- Maybe the resolved ESLint config that is used?

What else?

Norris1z:
This is a basic implementation of the `--verbose` flag which is a fix #188

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,7 +121,8 @@ module.exports.lintFiles = (patterns, options) => {
 };
 
 module.exports.debugInformation = options => {
-	const extensions = [...optionsManager.getExtensions(options).extensions];
+	const extensionOptions = optionsManager.getExtensions(options);
+	const extensions = extensionOptions.extensions.concat(extensionOptions.extension);
 	console.log('EXTENSIONS: ' + extensions.join(' | '));
 
 	const ignores = optionsManager.getIgnores(options).ignore; // Display only files ignored by the user
--- a/index.js
+++ b/index.js
@@ -120,6 +120,20 @@ module.exports.lintFiles = (patterns, options) => {
 	});
 };
 
+module.exports.debugInformation = options => {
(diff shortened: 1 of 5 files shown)


sindresorhus:
This seems to be missing:

> * Maybe the resolved ESLint config that is used?

sindresorhus:
Instead of reimplementing the logic, I think it would be better to just put the logging right next to where the different things are created, for example, the resolved config could be logged where XO finishes resolving the config.

index.js
```diff
@@ -120,6 +120,35 @@ module.exports.lintFiles = (patterns, options) => {
...
 
+module.exports.debugInformation = options => {
+	const extensionOptions = optionsManager.getExtensions(options);
+	const extensions = extensionOptions.extensions.concat(extensionOptions.extension || []);
+	console.log('EXTENSIONS: ' + extensions.join(' | '));
+
+	const ignores = optionsManager.getIgnores(options).ignore; // Display only files ignored by the user
+	if (ignores !== undefined) {
+		globby(ignores).then(files => {
+			files.forEach(file => {
```
sindresorhus: Use a `for-of` loop.

Norris1z:
@sindresorhus ignores in the `.gitignore` file is handled by `globby` and are not added to the `options.ignores` array. Should i implement the logic to list ignored paths/files in the `.gitignore`?

Norris1z:
The travis builds passed but i don't know why this is still stuck at `Some checks haven’t completed yet`

fregante:
Given the age I'm going to close this PR, but it's definitely still a welcome change. Please open a new PR if you intend to work on it again ☺️

PR #366, opened 2019-01-13
Throw an error if run on an ignored file

### LINKED ISSUE #238: Show error if run on a single file that's ignored
When using xo on a single file no errors are shown if said file is ignored in xo's settings.

```
➜  agenda git:(master) DEBUG=xo xo test/fixtures/addTests.js 
  xo Using local install of XO. +0ms
➜  agenda git:(master) mv test/fixtures/addTests.js .
➜  agenda git:(master) ✗ DEBUG=xo xo ./addTests.js            
  xo Using local install of XO. +0ms

  addTests.js:1:1
  ✖   1:1   Filename is not in kebab case. Rename it to add-tests.js.     unicorn/filename-case
  ✖   2:3   Strings must use singlequote.                                 quotes
  ✖   2:3   Extra space after key none.                                   key-spacing
```

Norris1z:
This PR fixes #238 and it throws an error if you call xo on a single file which is ignored.


Norris1z:
@sindresorhus should it throw an error or a console.error with a process.exit

sindresorhus:
This change needs tests.

sindresorhus:
You're missing https://github.com/xojs/xo/issues/238#issuecomment-314194430

sindresorhus:
The `ignore` option is also not the only way a file could be ignored. There's also the default ignore patterns and `.gitignore`.

cli-main.js
```diff
@@ -138,6 +139,22 @@ const log = report => {
 	process.exit(report.errorCount === 0 ? 0 : 1);
 };
 
+const throwIfFileIsIgnored = file => {
+	if (input.includes(file)) {
+		throw new Error('You cannot run xo on an ignored file');
```
sindresorhus: The error should mention the file path.

sindresorhus:
> should it throw an error or a console.error with a process.exit

`console.error` and `process.exit`

Norris1z:
@sindresorhus I had to modify a test to get this to work because this pr changes the behaviour of that test. 
`ignore explicit files when in .gitgnore` was asserting that explicit files ignored in the `.gitignore` will be ignored silently whereas this pr outputs an error.

Also i think https://github.com/xojs/xo/issues/238#issuecomment-314194430 has already been taken care of because the input array holds all files that match the `glob` pattern when a `glob` is specified rather than a direct file.

fregante:
Given the age I'm going to close this PR, but it's definitely still a welcome change. Please open a new PR if you intend to work on it again ☺️

PR #372, opened 2019-01-20
 Turn all paths `cwd`-relative before linting

### LINKED ISSUE #223: Overrides not working when calling xo with absolute paths

futpib:
fix #223

### DIFF
--- a/index.js
+++ b/index.js
@@ -98,6 +98,8 @@ module.exports.lintFiles = (patterns, options) => {
 			cwd: options.cwd
 		}
 	).then(paths => {
+		paths = paths.map(x => path.relative(options.cwd, path.resolve(options.cwd, x)));
+
 		// Filter out unwanted file extensions
 		// For silly users that don't specify an extension in the glob pattern
 		if (!isEmptyPatterns) {
--- a/test/cli-main.js
+++ b/test/cli-main.js
@@ -51,13 +51,13 @@ test('overrides work with relative path', async t => {
 	await t.notThrowsAsync(main([file], {cwd}));
 });
 
-test.failing('overrides work with relative path starting with `./`', async t => {
+test('overrides work with relative path starting with `./`', async t => {
 	const cwd = path.join(__dirname, 'fixtures/overrides');
 	const file = '.' + path.sep + path.join('test', 'bar.js');
 	await t.notThrowsAsync(main([file], {cwd}));
 });
 
-test.failing('overrides work with absolute path', async t => {
+test('overrides work with absolute path', async t => {
 	const cwd = path.join(__dirname, 'fixtures/overrides');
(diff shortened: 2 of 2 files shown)


sindresorhus:
It feels weird to turn paths relative. What if the overrides specify absolute paths? Wouldn't it make more sense to make the overrides paths absolute instead?

futpib:
> What if the overrides specify absolute paths?

Should overrides even support absolute paths? If you put an absolute path override in you `package.json` it probably won't work on anyone else's setup or in CI. For instance `/Users/zanona/Developer/my-app/packages/shop-app/**` (from https://github.com/xojs/xo/issues/223#issuecomment-354548646) clearly won't work for anyone else (username is in the path, the path is Mac-specific, etc.)

> Wouldn't it make more sense to make the overrides paths absolute instead?

Would be hard to do this correctly since override paths are `minimatch` globs.

sindresorhus:
That's a good point.

sindresorhus:
You need to add this PR to https://issuehunt.io/r/xojs/xo/issues/223 to get the bounty.

niftylettuce:
@sindresorhus 	Can you please release a new version to NPM?  The last release was January and this fix resolves issues with people using `lint-staged`.

okanthony:
@sindresorhus ditto ^^ 🙏 

@niftylettuce in the meantime you could always use a git URL to point to `master` in your `package.json`

PR #389, opened 2019-06-12
Add webpack plugin to the readme

nstanard:
I've created a basic webpack plugin for XO to be used. This is brand new - I was having trouble with the loader one and wanted it to be simpler to install and use. I have plans to add more features as well - but I wanted to get this link up on XO.


nstanard:
@sindresorhus Hey! This could be an easy PR to review. Please and thank you. ;)

PR #405, opened 2019-10-07
Fix compatibility with being run in a child process

OmgImAlexis:
This let's the process exit in the same way but still allows the streams to finish before `child_process` kills it.

Ref: https://github.com/nodejs/node/issues/12921

### DIFF
--- a/cli-main.js
+++ b/cli-main.js
@@ -132,7 +132,7 @@ if (typeof options.space === 'string') {
 const log = report => {
 	const reporter = options.reporter ? xo.getFormatter(options.reporter) : formatterPretty;
 	process.stdout.write(reporter(report.results));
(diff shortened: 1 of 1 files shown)


sindresorhus:
The linked issue is old and might not be the reality today. What exactly does this PR fix for you? Make sure you test with the latest Node.js 12 version.

OmgImAlexis:
The issue is `process.exit()` doesn't wait for stdout to finish.

Ref: https://nodejs.org/api/process.html#process_process_exit_code

> Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

OmgImAlexis:
@sindresorhus any chance this can be merged?

PR #409, opened 2019-11-29
re-enable the `no-use-extend-native` plugin

christyh11:
poop


sindresorhus:
This was already done a long time ago.

PR #413, opened 2019-12-13
Replace singular Lodash packages with the main one

Richienb:
The use of Lodash Per-Method packages are [discouraged](https://www.npmjs.com/browse/keyword/lodash-modularized) and will be removed in [v5](https://github.com/lodash/lodash/issues/3838#issuecomment-398592530). Let's future proof now.


index.js
```diff
@@ -2,7 +2,7 @@
 const path = require('path');
 const eslint = require('eslint');
 const globby = require('globby');
-const isEqual = require('lodash.isequal');
+const isEqual = require('lodash/isEqual');
```
sindresorhus:
```suggestion
const {isEqual} = require('lodash');
```

index.js
```diff
@@ -45,7 +45,7 @@ module.exports.lintText = (string, options) => {
 	options = optionsManager.preprocess(options);
 
 	if (options.overrides && options.overrides.length > 0) {
-		const {overrides} = options;
+		const { overrides } = options;
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
Thanks :)

PR #414, opened 2019-12-28
Added nodejs13 support on travis

smondal:
Added travis build against Nodejs 13

Nodejs 13.0.0 has released on 10-10-2019
https://nodejs.org/en/download/releases/


sindresorhus:
Thanks for the PR, but I generally only test against LTS Node.js versions as the unstable versions break CI all the time.

Issue #330, opened 2018-06-05
Support an external config file instead of just package.json config


j0hnm4r5:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/xojs/xo/issues/330)

This is somewhat of a similar ask to #250, but with a little different use context: 

I'd like to put my XO configuration in a separate file (`.xorc.*`), just like how [ESLint](https://eslint.org/docs/user-guide/configuring#using-configuration-files) and [Prettier](https://prettier.io/docs/en/configuration.html) do it.

Having all of those config options in a single file really clutters up my `package.json`, and I'd like to be able to more easily copy/symlink files from a central location, instead of having to copy/paste configurations into my `package.json` every time.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#425 Find XO config based on linted file path](https://issuehunt.io/r/xojs/xo/pull/425)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/xojs/xo/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

j0hnm4r5:
That's beside the point though — XO is so customizable, that it's really frustrating that I _have to_ put my config in the `package.json`, especially when so many other tools make that optional. While it may work great for a lot of people, I'm personally not a fan of defining my configurations in my `package.json` — I want nothing in there besides metadata and dependencies.

I issued a [pull request](https://github.com/xojs/xo/pull/331) yesterday (hopefully) with the fix.

pvdlg:
A reason to adopt `cosmiconfig` would be https://github.com/xojs/xo/pull/279#issuecomment-356021126

sindresorhus:
This is now accepted. See https://github.com/xojs/xo/pull/279#issuecomment-407250935.

IssuehuntBot:
[...]

PR #331, opened 2018-06-05
Use cosmiconfig for config variety

j0hnm4r5:
Fixes #330

PR #420, opened 2020-02-12
Change configuration loading to cosmiconfig

bunysae:
Changed configuration loading to cosmiconfig module. As @sindresorhus [mention](https://github.com/xojs/xo/pull/279#issuecomment-407250935) YAML is disabled.
Fixes issue #330 and superseded PR #331 as this seems no longer maintained.


lib/options-manager.js
```diff
@@ -155,14 +156,32 @@ const normalizeOptions = options => {
...
+	const moduleName = 'xo';
+	// Yaml is disabled
+	const searchPlaces = ['package.json',
+		`.${moduleName}rc`,
+		`.${moduleName}rc.json`,
+		`.${moduleName}rc.js`,
+		`${moduleName}.config.js`];
+	const loaders = {noExt: defaultLoaders['.json']};
+	// Change working directory for load config
+	process.chdir(options.cwd);
```
pvdlg: Instead of changing the working directory directory, pass `options.cwd` to comisconfig in option `searchFrom`

sindresorhus:
Thanks for working on this. It would be good with some more tests. And it also needs to be documented in the readme.

lib/options-manager.js
bunysae: Done. I didn't saw this option before.

bunysae:
I added tests for every variety and changed the docs with my latest commit.

sindresorhus:
Sorry @bunysae... I went with https://github.com/xojs/xo/pull/425 as it was much more complete.

PR #422, opened 2020-02-13
Use new eslint-config-prettier override

pvdlg:
Use the new plugin specific override from https://github.com/prettier/eslint-config-prettier


sindresorhus:
This is not a breaking change, right?

pvdlg:
No

PR #423, opened 2020-02-13
Disable rules based on Node.js version

### LINKED ISSUE #421: Fix the `nodeVersion` tests
We need a new fixture for those. See: https://github.com/xojs/xo/commit/d7fa8752dbd0b79d5429b89a718d73a79d59a4d8# and other disabled tests in that commit.

pvdlg:
This PR change how the internal logic works regarding the `nodeVersion`. Instead of disabling rules by default and enabling them based on the version we now do the opposite.

That allow us to define whatever rules we like regardless of the Node version and then disable some based on the user's node version.

This way we won't have to change anything when we increase the minimum Node version supported by XO !

For XO users it doesn't change anything.

Fix #421

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -43,12 +43,12 @@ const DEFAULT_CONFIG = {
 };
 
 /**
- * Define the rules that are enabled only for specific version of Node.js based on `engines.node` in package.json or the `node-version` option.
+ * Define the rules config that are overwritten only for specific version of Node.js based on `engines.node` in package.json or the `node-version` option.
  *
  * The keys are rule names and the values are an Object with a valid semver (`4.0.0` is valid `4` is not) as keys and the rule configuration as values.
  *
- * Each entry define the rule config and the minimum Node.js version for which to set it.
- * The entry with the highest version that is compliant with the `engines.node`/`node-version` range will be used.
+ * Each entry define the rule config and the maximum Node.js version for which to set it.
+ * The entry with the lowest version that is compliant with the `engines.node`/`node-version` range will be used.
  *
  * @type {Object}
  *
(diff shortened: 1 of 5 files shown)


sindresorhus:
Good idea! This is much better.

lib/options-manager.js
```diff
@@ -66,26 +66,50 @@ const DEFAULT_CONFIG = {
...
+		'10.0.0': 'off'
+	},
+	'node/prefer-global/url': {
+		'10.0.0': 'off'
+	},
+	'no-useless-catch': {
+		'10.0.0': 'off'
+	},
+	'prefer-named-capture-group': {
+		'10.0.0': 'off'
```
sindresorhus: I noticed now that this no longer reports when I run XO on https://github.com/avajs/ava (It does with XO 0.26.0)

Its `node` version target is: `"node": ">=10.18.0 <11 || >=12.14.0 <13 || >=13.5.0"`

pvdlg: I just tested that with `>=10.18.0 <11 || >=12.14.0 <13 || >=13.5.0` the current code turn off only the following rules:
  - `node/prefer-global/text-encoder`
  - `node/prefer-global/text-decoder`
  - `unicorn/prefer-flat-map`
  - `node/prefer-promises/dns`
  - `node/prefer-promises/fs`

Maybe you have an outdated version of `eslint-config-xo` so the rule is not enabled in the first place?

pvdlg: I also tried to run xo 0.26.0 on the current codebase of https://github.com/avajs/ava and the rule doesn't report.

sindresorhus: Hmm, weird. I couldn't reproduce it multiple times earlier. Just tried reinstalling node_modules everywhere and now it of course works... Sorry for wasting your time.

pvdlg: No problem! I'd rather spend more time to make sure it works.
Sometimes weird stuff happens with package-lock.json......

PR #426, opened 2020-02-17
Add TypeScript support

### LINKED ISSUE #373: Built-in support for TypeScript
[Issuehunt badges](https://issuehunt.io/r/xojs/xo/issues/373)

Now that the [TypeScript project is officially committed to using and supporting ESLint](https://eslint.org/blog/2019/01/future-typescript-eslint), I would like to explore adding built-in support for TypeScript in XO. You can already use TypeScript with XO, but I would like to make it *just work*.

I have already updated the shareable TS config for the new plugin: https://github.com/xojs/eslint-config-xo-typescript/commit/1161b418bebab5511d8dddf4cba98a71bb1311e5

XO will add TypeScript and the TypeScript ESLint plugin as a dependency.

We still need to discuss how it will exactly work. Can we manage to do it without any setting?

Note to self: Should lint `foo.d.ts` (TS definition files) in non-TS projects automatically.

---

This issue requires you to have extensive JS knowledge. It's not an easy issue. Only undertake it if you are confident you can handle it and you intend to see it through.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $200.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($200.00)

### Submitted pull Requests
- [#426 Add TypeScript support](https://issuehunt.io/r/xojs/xo/pull/426)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/xojs/xo/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

pvdlg:
Fix #373
Fix xojs/eslint-config-xo-typescript#15

Xo will now automatically lint `.ts` and `.tsx` with the config from `eslint-config-xo-typescript`. No configuration needed!

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#373: Built-in support for TypeScript](https://issuehunt.io/repos/40053602/issues/373)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>


sindresorhus:
Can you make it handle `.d.ts` too?

Per issue description:

> Should lint foo.d.ts (TS definition files) in non-TS projects automatically.

I don't have that many actual TypeScript projects, but I have hundreds of JS packages with a `*.d.ts` file and it would be nice to be able to enforce best practices in those.

pvdlg:
Can you make it handle .d.ts too?

It should lint `.d.ts` files. `index.d.ts` is considered to have the extension `.ts` so it would be included.
Did you encountered an issue linting a `.d.ts` file?

sindresorhus:
Good point. I was out on my phone and didn't check yet.

sindresorhus:
There's a problem with `npm link`'ing XO. I tried `npm link` in XO and then `npm link xo` in, for example, https://github.com/sindresorhus/make-dir, and I'm getting:

```
Error: Failed to load config "xo-typescript" to extend from.
Referenced from: BaseConfig
    at configMissingError (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:265:9)
    at ConfigArrayFactory._loadExtendedShareableConfig (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:826:23)
    at ConfigArrayFactory._loadExtends (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:731:25)
    at ConfigArrayFactory._normalizeObjectConfigDataBody (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:660:25)
    at _normalizeObjectConfigDataBody.next (<anonymous>)
    at ConfigArrayFactory._normalizeObjectConfigData (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:596:20)
    at _normalizeObjectConfigData.next (<anonymous>)
    at createConfigArray (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:340:25)
    at ConfigArrayFactory.create (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/config-array-factory.js:395:16)
    at createBaseConfigArray (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/cli-engine/cascading-config-array-factory.js:86:48)
```

It works if I manually copy the XO directory into `node_modules` of `make-dir`.

sindresorhus:
> It works if I manually copy the XO directory into node_modules of make-dir.

However, then there are some parse errors:

```
~/dev/oss/make-dir master
❯ xo

  index.test-d.ts:2:16
  ✖  2:16  Parsing error: Unexpected token =

  index.d.ts:4:9
  ✖  4:9   Parsing error: Unexpected token namespace

  2 errors
```

sindresorhus:
There's also a problem with reading `tsconfig.json` directly. `tsconfig.json` supports comments and it's very common to put it in there. I would look into how `@typescript-eslint/eslint-plugin` reads the tsconfig.

Here's the error:

```
~/dev/oss/got
❯ xo
JSONError: JSON Error in /Users/sindresorhus/dev/oss/got/tsconfig.json:

  3 | 	"compilerOptions": {
  4 | 		"outDir": "dist",
> 5 | 		"target": "es2018", // Node.js 10
    | 		                    ^
  6 | 		"lib": [
  7 | 			"es2018"
  8 | 		]
```

readme.md
```diff
@@ -25,8 +25,9 @@ Uses [ESLint](https://eslint.org) underneath, so issues regarding rules should b
 - Beautiful output.
 - Zero-config, but [configurable when needed](#config).
 - Enforces readable code, because you read more code than you write.
-- No need to specify file paths to lint as it lints all JS files except for [commonly ignored paths](#ignores).
+- No need to specify file paths to lint as it lints all JS/TS files except for [commonly ignored paths](#ignores).
 - [Config overrides per files/globs.](#config-overrides)
+- TypeScript supported by default
```
sindresorhus: Make linkify this to the TypeScript section?

pvdlg:
The loading is probably related to how ESlint resolve config/plugin names. I'll look into it to see if something can done (maybe passing absolute path to ESlint?)

I'll check the parsing errors and the comment in tsconfig.json tomorrow.

sindresorhus:
Non-TS projects with a `d.ts` file don't usually have a `tsconfig.json` file, but TS is way too loose by default. We should include some strict defaults for such cases, by using https://github.com/sindresorhus/tsconfig/blob/master/tsconfig.json

pvdlg:
> Non-TS projects with a d.ts file don't usually have a tsconfig.json file, but TS is way too loose by default. We should include some strict defaults for such cases, by using https://github.com/sindresorhus/tsconfig/blob/master/tsconfig.json

I'm not sure it's possible. We look at the `tsconfig.json` only to figure out the path and pass it to [parserOptions.project](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser#parseroptionsproject) but it's the parser that loads it. It doesn't seems to be a way to pass the config to the parser programatically.

sindresorhus:
> I'm not sure it's possible. We look at the tsconfig.json only to figure out the path and pass it to parserOptions.project  but it's the parser that loads it. It doesn't seems to be a way to pass the config to the parser programatically.

https://github.com/typescript-eslint/typescript-eslint/issues/1613

pvdlg:
I can't reproduce the issue you are having on https://github.com/sindresorhus/make-dir

I cloned the repo, then `npm install -D xojs/xo#typesscript`. I also had to add `typescript` to the project with `npm install typescript -D` which is a required dependency of the parser it seems.
Finally I had to add a `tsconfig.json` with:
```json
{
  "include": ["**/*.ts"]
}
```
Then running xo:
```
~/g/make-dir (master ✚1…1) ❯ xo                                                                                                                 264ms

  index.test-d.ts:11:38
  ✖  11:38  Use octal literals instead of parseInt().  prefer-numeric-literals
  ✖  20:42  Use octal literals instead of parseInt().  prefer-numeric-literals

  2 errors
```

It seems that a `tsconfig.json` is required by `@typescript-eslint/parser`. From the doc:
> This setting is required if you want to use rules which require type information

Maybe we can try to figure out which rules require `parserOptions.project` and disable them if no `tsconfig.json` is found. But that would be a pain to maintain as every time we add a rule to `eslint-config-xo-typescript` we would have to disable it in XO when no `tsconfig.json` is found.

Another solution is to enable the Typescript rules only when a `tsconfig.json` is found even if the file is a `.ts`. Which would require projects with only a `.d.ts` file to add a `tsconfig.json`.

sindresorhus:
> Maybe we can try to figure out which rules require parserOptions.project and disable them if no tsconfig.json is found. But that would be a pain to maintain as every time we add a rule to eslint-config-xo-typescript we would have to disable it in XO when no tsconfig.json is found.

Yeah, I don't think that's a realistic solution.

> Another solution is to enable the Typescript rules only when a tsconfig.json is found even if the file is a .ts. Which would require projects with only a .d.ts file to add a tsconfig.json.

That would severely limit the usefulness of this. There are much more projects with `.d.ts` than there are pure TS projects and I don't want to have to add a tsconfig to pure JS projects.

sindresorhus:
Hmm, an idea. We could just write out a tsconfig file to a temp directory and point `parserOptions.project` to that.

pvdlg:
For info https://github.com/typescript-eslint/typescript-eslint/issues/406 which is essentially the same as the one you opened hasn't been answered.

Also see https://github.com/typescript-eslint/typescript-eslint/issues/689. It seems the tsconfig.json file is used at a minimum to read the `include` property to figure out all the files that are part of the project and compile them.

To be honest I though a `tsconfig.json` was mandatory for every TypeScript project and I didn't think about JS project have a type definition file.

If we add a default `tsconfig.json` I'm not sure if we should set the `include` with just the file we are currently linting or with `["**/*.ts",**/*.tsx"]`. In the first case we wouldn't lint much as we would loose the ability to use the type definition from other files. In the other case we might create unintended behaviors on some projects.

sindresorhus:
> If we add a default tsconfig.json I'm not sure if we should set the include with just the file we are currently linting or with ["**/*.ts",**/*.tsx"]. In the first case we wouldn't lint much as we would loose the ability to use the type definition from other files. In the other case we might create unintended behaviors on some projects.

We should only do it for `.d.ts` files. Actual TypeScript projects are expected to have their own `tsconfig` file.

sindresorhus:
Hmm, maybe we could fix https://github.com/xojs/eslint-config-xo-typescript/issues/15 at the same time (It's a huge pain-point). Often the tsconfig file contains more files than you actually want to lint. We could read the tsconfig file and switch out `includes` with only the files being linted.

// @fregante

pvdlg:
After doing a lot of tests, we can create a temporary one and pass it to `parserOptions.projet`. However the temporary `tsconfig.json` has to contains absolute paths in the `include` property.

This will **not** work:
```json
{
  "include": ["**/*.ts"]
}
```

This will work:
```json
{
  "include": ["/users/pvdlg/project/**/*.ts", "/users/pvdlg/project/**/*.tsx"]
}
```

So we could do:
- if we don't find a `tsconfig.json` => create a temp one with `"include": ["${options.cwd}/**/*.ts", "${options.cwd}/**/*.tsx"]`
- if we find a `tsconfig.json` => create a temp one with `"include": ["${options.cwd}/**/*.ts", "${options.cwd}/**/*.tsx"]` that extends the file we've found

I think it should solve both xojs/eslint-config-xo-typescript#15 and projects without a `tsconfig.json`.

That seems like a clunky workaround, but it's the only solution I've found.

sindresorhus:
> if we don't find a tsconfig.json => create a temp one with "include": ["${options.cwd}/**/*.ts", "${options.cwd}/**/*.tsx"]

I'm not sure how `@typescript-eslint/parser` works under the hood with the `include` option, but just globbing everything might be slow. Wouldn't it be better to pass in the `.ts` files XO has already found?

Also, here we should inject some sensible defaults: https://github.com/sindresorhus/tsconfig/blob/9c49b210a7e7dee5a0feeb8433c8cc165f6319bb/tsconfig.json#L15-L19

> That seems like a clunky workaround, but it's the only solution I've found.

I'm totally fine with it being clunky if it fixes the problem.

pvdlg:
@sindresorhus I updated the PR per our discussion.
This should solve the case of projects without a `tsconfig.json` and xojs/eslint-config-xo-typescript#15.

Note that in the case of project without a `tsconfig.json`,  when linting a subset of files `@typescript-eslint/eslint-plugin` will have no way to know about other files in the project and that might affect how the rules behave.
For example if you have a project with files `src/a.ts` and `src/b.ts` and run `xo src/a.ts` `@typescript-eslint/eslint-plugin` will execute its rule without knowledge of `b.ts`. Not sure if that's really an issue though. Plus there is nothing we can do other than globing the whole project...

This problem will not happen if you have a `tsconfig.json`.

lib/options-manager.js
```diff
@@ -73,47 +87,89 @@ const mergeWithFileConfig = options => {
...
+
 	return {options, prettierOptions};
 };
 
+const makeTSConfig = (tsConfig, tsConfigPath, files) => {
+	const config = {files: files.filter(isTypescript)};
+
+	if (tsConfigPath) {
+		config.extends = tsConfigPath;
+		config.include = arrify(tsConfig.include).map(pattern => toAbsoluteGlob(pattern, {cwd: path.dirname(tsConfigPath)}));
```
sindresorhus: Can you also ensure it's strict and apply this config:

```
"target": "es2018", // Node.js 10
"newLine": "lf",
"strict": true,
"noImplicitReturns": true,
"noUnusedLocals": true,
"noUnusedParameters": true,
"noFallthroughCasesInSwitch": true,
```

?

sindresorhus:
Can you also ensure the generated tsconfig is strict by adding this config to it?

```
"target": "es2018", // Node.js 10
"newLine": "lf",
"strict": true,
"noImplicitReturns": true,
"noUnusedLocals": true,
"noUnusedParameters": true,
"noFallthroughCasesInSwitch": true,
```

sindresorhus:
Can you mention in the readme that it works for `d.ts` files in JS projects without a tsconfig file? I don't think that would be obvious to the user.

sindresorhus:
I've been testing this on a bunch of projects and it seems to work great. 👌

lib/options-manager.js
```diff
@@ -322,6 +383,27 @@ const mergeWithPrettierConfig = (options, prettierOptions) => {
 	);
 };
 
+const buildTSConfig = options => config => {
+	if (options.ts) {
+		config.baseConfig.extends = config.baseConfig.extends.concat('xo-typescript');
+		config.baseConfig.parser = '@typescript-eslint/parser';
```
sindresorhus: This is already defined by `eslint-config-xo-typescript`.

And there we resolve it to its absolute path, not sure if that makes a difference though: https://github.com/xojs/eslint-config-xo-typescript/blob/99c4590d599933f754c498c02cf95cec3c6d04a1/index.js#L3

pvdlg:
I made the requested changes.

I added the default config only when no `tsconfig.json` is found.
I don't think it's a good idea to add those when the project has a `tsconfig.json` because it would result in the linter using different options than the the compiler. Also there is a lot of project that doesn't set options like `strict`, `noImplicitReturns` and use the default on purpose. For example when transitioning from JS to TS. What do you think?

sindresorhus:
Oh yeah. Definitely. I meant when there’s no tsconfjg file.

lib/options-manager.js
pvdlg: Done

sindresorhus:
Manually tested it again and seems to work great.

sindresorhus:
Woooot! Really awesome that you decided to implement this. 🎉✨

Should I do a new release?

bmealhouse:
Very excited for these changes. Thanks so much @pvdlg!

pvdlg:
> Should I do a new release?

Yes that would be awesome!

sindresorhus:
It's out: https://github.com/xojs/xo/releases/tag/v0.27.0

PR #431, opened 2020-02-24
Various fixes to TypeScript integration

pvdlg:
Fix an issues introduced in #426

The key used to cache and group config wasn't working properly and  had `ts` always be `undefined`, preventing to split JS and TS files in two groups.

As we were generating the temporary `tsconfig.json` with a different name every run the ESLint cache wasn't working properly has it would determine the config changed in between runs (due to `parserOptions.project`) being different.
We now generate those files in the XO cache with a consistent name, so in each run, the same set of files/tsconfig.json would result in the same temporary tsconfig.json file name.

Finally this solve a very rare case with a project with multiple XO configs. In such case were generating one temporary tsconfig.json per XO config. We now correctly generate temporary tsconfig per original one (and just one if the project doesn't have a tsconfig).
That would have created a situation were `@typescript-eslint/eslint-plugin` would have been unaware of the TS files that are not part of the currently linted group.

**Important for the release note**:
After updating XO to a version with this fix, users will need to delete the current XO cache:
```
$ rm node_modules/.cache/xo
```

That's because now `node_modules/.cache/xo` should be a directory instead of a file so we can save both the ESLint cache and the temporary tsconfig.json.


lib/options-manager.js
```diff
@@ -179,9 +179,9 @@ const mergeWithFileConfigs = async (files, options) => {
 };
 
 /**
- * Generate a unique and consitent path for the temporary `tsconfig.json`.
```
pvdlg: Just out of curiosity, what's the reason for changing the function comment format?
[The JSDoc](https://jsdoc.app/about-getting-started.html) use the format with each intermediate line starting with a `*`.

pvdlg: btw I didn't noticed you have changed the other comments, otherwise I would used your preferred format here.

sindresorhus: I think it's just needless noise and it makes it harder to edit things, especially annoying to add multi-line code blocks in the comment.

pvdlg: Make sense. My IDE format it with the `*` automatically and it when I insert a new line so I never really paid much attention.
I'll use your preferred format in future PRs!

sindresorhus: Don't worry. I didn't want to bother you with nitpick style issues. I'm just very particular about that. Hence why I made XO in the first place :P

sindresorhus:
I encountered an issue with the new TS integration. Clone this branch: https://github.com/sindresorhus/emittery/tree/upgrade-xo And `npm install && npm test`, and you'll get this error:

```
Error: Failed to load parser '/Users/sindresorhus/dev/oss/emittery/node_modules/@typescript-eslint/parser/dist/parser.js' declared in 'BaseConfig': Cannot find module 'typescript'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/Users/sindresorhus/dev/oss/emittery/node_modules/@typescript-eslint/typescript-estree/dist/parser.js:20:25)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
npm ERR! Test failed.  See above for more details.
```

pvdlg:
You need typescript in your project dependencies to be able to lint either `.ts` or `.d.ts` files.

I didn't add `typescript` as a dependency of XO because we don't want to install it for users that don't use typescript nor force a specific version to used.

I didn't think of the case of a project that doesn't have it in it's dependencies. Maybe I can add that to the readme.

sindresorhus:
XO is all about convenience though. I don't really want to install `typescript` manually in every JS project that has a `d.ts` (which is most of my JS projects these days). Couldn't we add it with a loose semver range? Then the user could specify their own more specific range if needed and it would be deduplicated to that. It could be `"typescript": "^3.0.0"`.

pvdlg:
See PR #434

That would mean every XO user issue will download typescript on `npm install` even if they don't use it. Personally I don't mind, disk space is cheap.

fregante:
Is `@typescript-eslint/parser` still required as additional dependency? I'm trying on Refined GitHub and it's complaining

pvdlg:
Try `rm -rf node_modules && rm package-lock.json && npm install`

fregante:
The usual then.

sindresorhus:
@pvdlg This change turned out to be really annoying because of Travis caching. Every time I upgrade XO, it will fail, and I have to manually clear Travis cache and then rerun.

How about we change the cache directory `.xo-cache/` => `.xo-cache2/` just for now, and sometime in the future we can change it back?

PR #432, opened 2020-02-25
Sort files path before generating unique cache key

pvdlg:
Follow up to #431

On large project the order of files passed to the `getTsConfigCachePath` can vary from one run to the other, which result in a different tsconfig temporary file name, invalidating the ESLint cache.

This PR applies a sort to the files list to maintain a consistent file name.

Sorry about all the different PRs to finally get that functionality right...


sindresorhus:
> Sorry about all the different PRs to finally get that functionality right...

Don't worry. It's a difficult feature.

PR #425, opened 2020-02-17
Find XO config based on linted file path

pvdlg:
fix #331, fix #279, fix #330, fix #420, fix #65

This change the way to load the XO config files. Instead of looking for the config in directly `cwd` we now search for a config file starting from the directory of each linted and by walking up parent directories.

The config files supported are now: `pacakge.json`, `.xorc` (JSON only), `.xorc.json`, `.xorc.js`, `xo.config.js`

This PR will help to implement #373 as we can now build a different config per file mode easily. That will allow us to apply the Typescript rules only to the `.ts`/`.tsx` files.

A few other things incidentally added by the PR:
- #65 is now fixed
- SamVerschueren/vscode-linter-xo#58 should be fixed without any changes in `vscode-linter-xo`
- `lintFiles` is now completely async allowing to parallelise looking for config files (async) with building/merginf config (sync)
- Reorganized the code a little bit for clarity as the files and function started to be quite large
- Replace `multimatch` with `micromatch` as it's the library used by `globby` which is already in the dependency tree

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#65: A way to override the default ignores](https://issuehunt.io/repos/40053602/issues/65)
- [#330: Support an external config file instead of just package.json config](https://issuehunt.io/repos/40053602/issues/330)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

PR #433, opened 2020-02-25
Clarify doc regarding `extension` option

pvdlg:
Ref https://github.com/xojs/xo/issues/428#issuecomment-591005878

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -244,9 +244,12 @@ Use one or more [shareable configs](https://eslint.org/docs/developer-guide/shar
 
 ### extensions
 
-Type: `string[]`
+Type: `string[]`\
+Default: `['js', 'jsx', 'ts', 'tsx']`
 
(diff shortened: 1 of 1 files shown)


pvdlg:
So that's a consequence of #425 that I didn't really forsee.
If this limitation is an issue I can change the code to:
- First glob for XO config files
- For each config file get the `extension` option
- Glob the files to lint, taking in account the extension config 
- Then call `mergeWithFileConfigs`

That would allow us to retain the ability to use `extensions` inside a config file rather than limiting that to CLI arg only. The trade-off is that we would glob 1 + n times (where n is the number of XO config found) instead off once. In the most common case (1 XO config) we would glob 2 times instead of once.

sindresorhus:
> That would allow us to retain the ability to use extensions inside a config file rather than limiting that to CLI arg only. 

Yeah, I think this would be the better and expected behavior. ESLint 7 will also allow specifying extensions in config files: https://eslint.org/blog/2020/02/whats-coming-in-eslint-7.0.0#file-extensions-in-config-files

PR #434, opened 2020-02-25
Add `typescript` as a dependency

pvdlg:
Per https://github.com/xojs/xo/pull/431#issuecomment-591052196


fregante:
# 🎈 

https://packagephobia.now.sh/result?p=xo

PR #436, opened 2020-02-28
Get `extensions` and `ignores` from config files

pvdlg:
Fix #433


sindresorhus:
Looks good. Would you consider this a breaking change?

pvdlg:
Yes it's a breaking change if you compare `0.27.0`

PR #439, opened 2020-03-08
Add `unicorn/string-content` rule

sindresorhus:
See: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/string-content.md

Anything else useful we can enforce in strings?

// @pvdlg @fregante

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -113,6 +113,14 @@ module.exports = {
 				}
 			}
 		],
+		'unicorn/string-content': [
+			'error',
+			{
+				patterns: {
+					[/\.\.\./.source]: '…'
+				}
+			}
+		],
 
 		// The character class sorting is a bit buggy at the moment.
 		'unicorn/better-regex': [
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
 		"eslint-plugin-node": "^11.0.0",
 		"eslint-plugin-prettier": "^3.1.2",
 		"eslint-plugin-promise": "^4.2.1",
-		"eslint-plugin-unicorn": "^17.0.1",
+		"eslint-plugin-unicorn": "^17.2.0",
 		"find-cache-dir": "^3.0.0",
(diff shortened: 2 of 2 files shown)


fregante:
I’m unsure. Would this rule require exceptions every time `... on Commit` is used in a GraphQL query? Could it use tags to automatically determine code strings to skip?

```js
gql` repo() { ... on Commit {}}`
```

sindresorhus:
I think the rule should be excluded on `gql` by default as it's code in a string and not a pure string.

// @fisker

sindresorhus:
@fregante The rule has `'` => `’` built-in. Should we also do `"foo"` => `“foo”`? The complication is that the changed quotes are different characters for left and right.

sindresorhus:
We could maybe include some common misspellings? `node.js` => `Node.js`, `NodeJS` => `Node.js`?

fregante:
I think this rule is something not all projects can handle by default. CSS selectors have lots of `"` and `'`. Let's start with `'` for now

isghe:
It's really dangerous; it breaks a program I am working on:
from:
```
'readMessage(:queue, :id, \'ZCP\', :cursor)';
```
to
```
'readMessage(:queue, :id, ’ZCP’, :cursor)';
```
breaking a query to oracle db.

I think it shouldn't change the content of a string.

tusbar:
Agreed, this rule should be reverted as it breaks any string that’s not displayable text.

niftylettuce:
Agreed, same exact thing happened on my end that @isghe.

niftylettuce:
Please resolve this, I'm seeing stuff like this get replaced as well:

> `package-scripts.js` (nps)

```diff
-pretestMongo: "mongo forwardemail_test --eval 'db.dropDatabase()'",
+pretestMongo: 'mongo forwardemail_test --eval ’db.dropDatabase()’',
```

Issue #445, opened 2020-03-26
0.28.1 breaks support for space option in typescript


EdJoPaTo:
I assume #442 overrides the options set by space.

`npm i xo@0.28.0 && npx xo` works fine.
`npm i xo@0.28.1 && npx xo` does not.

sindresorhus:
// @pvdlg

pvdlg:
For Typescript files we set the rule `@typescript-eslint/indent` explicitly based on the XO `space` option: https://github.com/xojs/xo/blob/2e39794d5ad4017ed2545d9a43691d4c4b0f7ade/lib/options-manager.js#L285

So not surer what is the problem you are reporting.

Can you provide a reproductible case with your config, the file you lint and what report you get and expect?

EdJoPaTo:
create a temp folder, put in a ts file, for example index.ts with this content:
```ts
export function foo(): void {
  console.log('bar');
}
```

and a package.json with this content
```json
{
  "name": "xo-testing",
  "xo": {
    "space": 2
  }
}
```

Then run these two commands:
```sh
npm i xo@0.28.0 && npx xo
npm i xo@0.28.1 && npx xo
```


Using 0.28.0 this works fine without errors. 
0.28.1 reports for the rule 'indent' (which is only javascript specific, not for typescript) for expected tabs.

pvdlg:
Thanks! Will look into that tonight

pvdlg:
So the problem is the `indent` rule is set in the default XO config: https://github.com/xojs/eslint-config-xo/blob/06e4e1416bdbf2df56984adca8039683ebad735c/index.js#L305

So as a result for TS files we end up with both `indent` and `@typescript-eslint/indent` to be set up.

Fixing https://github.com/xojs/eslint-config-xo-typescript/issues/23 will solve the issue

pvdlg:
Never mind my last comment, the `indent` rule is not part of those.

PR #446, opened 2020-03-26
Support `semicolon` option with TypeScript

EdJoPaTo:
This was tested with v0.28.0 and works for me (`npm link`ed it into one of my projects).
It does not seem to work with v0.28.1. See #445


sindresorhus:
Thanks for looking into this. Would you be able to add a test?

EdJoPaTo:
I added tests for the options space and semicolon. As this branch is currently on 0.28.0 they still work.
Merging 0.28.1 into the branch results in the tests failing. (See #445)

lib/options-manager.js
```diff
@@ -296,7 +296,12 @@ const buildXOConfig = options => config => {
 	}
 
 	if (options.semicolon === false && !options.prettier) {
-		config.rules.semi = ['error', 'never'];
+		if (options.ts) {
```
pvdlg, line 299: This would cause the same issue as with the `indent` rule that you reported in #445.

This code should be:

```
if (options.semicolon === false && !options.prettier) {
  config.rules.semi = ['error', 'never'];
  config.rules['semi-spacing'] = ['error', {
    before: false,
    after: true
  }];

  if (options.ts) {
    config.rules['@typescript-eslint/semi'] = ['error', 'never'];
  }
}
```

EdJoPaTo: The semi rule should be disabled for typescript linting:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/semi.md

EdJoPaTo: But you're right, the base 'semi' rule has to be overridden

EdJoPaTo: Actually… why?
The semi rule is already set 'off' [here](https://github.com/xojs/eslint-config-xo-typescript/blob/e69953688059fabf2df8c29b18b8b8f74fe9f0f3/index.js#L407).

This does seem to work when the branch is not merged with the current master.
When merging it to the master the 'semi' rule is failing but it should be overridden to be 'off'.
Not sure what the reason for it is.

I used `npm link` on this branch and linked it into a small testing project with an index.ts and index.js
```js
export function foo(): void {
	console.log('bar')
}
```

As soon as I merge master into this branch the 'semi' rule appears for the index.ts which shouldn't occur. There is something else still wrong.

pvdlg: The problem is related to the order of `extends`. Currently `eslint-config-xo-typescript` is imported before `eslint-config-xo`.
I'll look at changing that.
I think ideally it should be: `eslint-config-xo`, then `eslint-config-xo-typescript` then users `extends`

EdJoPaTo: will be fixed with #453

EdJoPaTo:
As this works in combination with #453 this can be merged after merging #453

PR #453, opened 2020-04-02
Add TS extends after base XO extends and before user extends

pvdlg:
Per https://github.com/xojs/xo/pull/446#discussion_r402357614

This essentially revert #451 and #442 that weren't the correct fixes.

Instead we now add the typescript specific extends after the base XO config and before the user extends.
This allow `eslint-config-xo-typescript` to override some rules from `eslint-config-xo` and users to extend a config that override rules from both `eslint-config-xo-typescript` and `eslint-config-xo`.

@EdJoPaTo can you confirm this fix the issues reported?

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -259,9 +259,9 @@ Transform an XO options into ESLint compatible options:
 const buildConfig = (options, prettierOptions) =>
 	flow(
 		buildXOConfig(options),
+		buildTSConfig(options),
 		buildExtendsConfig(options),
-		buildPrettierConfig(options, prettierOptions),
-		buildTSConfig(options)
+		buildPrettierConfig(options, prettierOptions)
 	)(mergeWith(getEmptyOptions(), DEFAULT_CONFIG, normalizeOptions(options), mergeFn));
 
 const buildXOConfig = options => config => {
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
Awesome! Everything I tested worked as it should.

The only thing i noticed: xo is getting slower. But that's a different topic from being functional correct.

pvdlg:
> The only thing i noticed: xo is getting slower. But that's a different topic from being functional correct.

You mean with this particular change? Or with Typescript files? Or just overall?

EdJoPaTo:
In general. This change does not seem to have a significant effect on that.

I tested two repos of mine, one 2.2k LOC JavaScript, one 5.6k LOC TypeScript.
`npx xo; npx xo; time npx xo` (multiple times in order to make sure caching works) results in about 4.6 Seconds on my i3 desktop (daily driver) and about 2.0 seconds on my i7 laptop.
TypeScript or JavaScript and LOC does not seem to matter much here.

For both projects I tested xo@0.27 and xo@0.28 and got consistent timings (about 4.6s).
I tested xo@0.22 until xo@0.26 only with the JavaScript Repo and it took about 2.4s - 2.6s each time.
xo@0.20 and xo@0.21 took about 5.0s - 5.2s on the JavaScript repo. (Looking at 0.22 it introduces eslint 5 which might explain the increased speed in later versions)
xo@0.18 was even slower with 8s on the JavaScript repo (0.20 introduces eslint 4)

When considering devs always have their beefy machines and big monitors and users use smaller and cheaper machines I often try the UX on a raspberry pi. A pi 4 (which is already insane by itself) just lying around here resulted in 5.3 seconds for the TypeScript project and 3.9s for the JavaScript project (first time there is a significant difference between TS and JS and that's even faster than my daily driver desktop).

I remembered xo being faster but checking on that seems like it wasnt.
Still 4+ seconds to run xo isnt _fast_.

PR #462, opened 2020-04-11
Temporarily disable the `unicorn/string-content` rule

fregante:
Not all strings are user-facing so a blanket "replace all ... with …" can break all non-tagged strings.

As reported in https://github.com/xojs/xo/pull/439#issuecomment-598751860

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -113,14 +113,16 @@ module.exports = {
 				}
 			}
 		],
-		'unicorn/string-content': [
-			'error',
-			{
-				patterns: {
-					[/\.\.\./.source]: '…'
-				}
-			}
-		],
+
+		// TODO: Restore when it becomes safer
+		// 'unicorn/string-content': [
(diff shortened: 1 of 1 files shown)


fregante:
Actually I don't think this can be enabled in XO at all and https://github.com/sindresorhus/eslint-plugin-unicorn/issues/681 (whitelisting) would be a better user-facing solution.

sindresorhus:
I don't think this PR is needed as we changed the rule to include no patterns by default: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/08fad5552c64a386827e7f554fa629d85f5c5200

fregante:
Uhm? Yeah, that’s the rule itself, but XO does set the rule right here

sindresorhus:
I was confused which repo I was on.

sindresorhus:
Can you comment it out instead of removing it completely?

PR #470, opened 2020-05-02
ci: use GitHub Actions instead of Travis

pizzafox:
This changes the CI to use GitHub Actions instead of Travis.
It does a matrix of Node.js v10 & v12 on the `windows-latest` and `ubuntu-latest` OSs, which run on pushes and pull requests.
The matrix can be configured in the `jobs.test.strategy.matrix` section of the config located at `.github/workflows/ci.yml`.

Since this is using the Coveralls GitHub Action it also removes the need for the [`coveralls`](https://www.npmjs.com/package/coveralls) `devDependency` in CI, but I left that untouched in case it should be left around for local dev usage.


sindresorhus:
I appreciate the PR, but I don't plan to switch to GitHub Actions yet. I feel like GitHub Actions is too immature. It changes quite often, things break, etc. And the config format is too verbose. Travis still works fine. I plan to switch at some point, but not for at least another year.

PR #472, opened 2020-05-08
Enable `import/no-anonymous-default-export` and `import/no-named-default`

fregante:
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-anonymous-default-export.md
Reports:
```js
export default () => {}
export default class {}
export default function () {}
```

Function names help automatic imports by IDEs

---

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-default.md

Reports: 
```js
// message: Using exported name 'bar' as identifier for default export.
import { default as foo } from './foo.js';
import { default as foo, bar } from './foo.js';
```

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -182,6 +182,7 @@ module.exports = {
 		],
 		'import/no-absolute-path': 'error',
 		'import/no-anonymous-default-export': 'error',
+		'import/no-named-default': 'error',
 		'import/no-webpack-loader-syntax': 'error',
 		'import/no-self-import': 'error',
 
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -181,6 +181,7 @@ module.exports = {
 			}
 		],
 		'import/no-absolute-path': 'error',
+		'import/no-anonymous-default-export': 'error',
 		'import/no-webpack-loader-syntax': 'error',
 		'import/no-self-import': 'error',
 


sindresorhus:
> Function names help automatic imports by IDEs

You mean autocomplete? Why can't the IDEs just take the name from the filename?

fregante:
> You mean autocomplete? 

Yeah, type `stripW|` and the IDE adds:

```js
import stripeWhitespace from './stripe-whitespace';

stripeWhitespace|
``` 

> Why can't the IDEs just take the name from the filename?
 
Don’t ask me, I don’t make IDEs 😅

sindresorhus:
I know. But I'm saying you're trying to fix it in the wrong place. This should instead be an issue on the IDE you use. I don't like having to have needlessly more verbose code just to please an IDE.

fregante:
It's not just an IDE issue, your functions will also appear as anonymous when debugging them (named `default` or `module.exports`)

sindresorhus:
So the debugger needs to be improved too then.

(I’m not saying I won’t accept this. Just something to consider.)

fregante:
> So the debugger needs to be improved too then

The debugger shows the real function name which as per spec is `default` or `module.exports`, debuggers can't change that.

> Which debugger are we talking about?

I think every debugger tries to show the function name in the stack, but yes users will see the file name as `at default (./node_modules/throw-error/index.js)` or `at module.require (./throw-error.js)`

With sourcemap enabled even bundlers will show the correct filename in this case. Rollup is the smartest one and the function appears as `const x = () => {throw Error()}` so it will have the user’s name.

I think however it's worth adding a name to an important function (so important in fact that it deserves its own file/module). 

For example this can also serve as a capitalization suggestion, no IDE will ever suggest importing the `url-search-params` module as `URLSearchParams`. If the module has `export default class URLSearchParams` this name is exactly what will be used, without extra guesses.

sindresorhus:
I'm still lukewarm about it. It does help debugging, but it also makes exports less elegant.

fregante:
Some will complain that this lengthens exports that aren’t functions or classes I’m sure, since `export default 42` becomes `const life = 42; export default life`

If that’s too annoying , it looks like the rule can be configured, but I guess the same principle applies in that case too: the export now has a name

jimmywarting:
Didn't like this rule... too opinionated. IDE should become smarter instead.

PR #487, opened 2020-06-15
The .gitattributes cause media/logo.png to have it's line endings modified

bjornstar:
At least when I check out the repo under WSL.


sindresorhus:
Not sure what the issue might be, but `* text=auto eol=lf` says that the `lf` normalization should only apply to text. Might be a case of an outdated Git version.

PR #508, opened 2020-11-22
Disable warnings when running in a GitHub Actions workflow

sindresorhus:


readme.md
```diff
@@ -292,6 +292,10 @@ XO will handle the [@typescript-eslint/parser `project` option](https://github.c
 
 See [eslint-config-xo-flow#use-with-xo](https://github.com/xojs/eslint-config-xo-flow#use-with-xo)
 
+## GitHub Actions
+
+XO uses a different formatter when running in a GitHub Actions workflow to be able to get inline annotations. XO also disables warnings here.
```
sindresorhus, line 297: @fregante Do you know any page I could link to with more info about the inline annotations?

fregante: Found these on their blog

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-status-checks#checks
https://developer.github.com/changes/2019-09-06-more-check-annotations-shown-in-files-changed-tab/

fregante:
I think CI would be the best place to show warnings, the problem is that they trigger annotations. If it were easy, I'd still have it display warnings without triggering annotations, but IMHO this is an acceptable stopgap.

PR #513, opened 2020-12-28
Update dependencies

fregante:
Globby is 2 versions behind, I'm gonna try to update it.

For now, here's every other dependency updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,15 +51,15 @@
 		"typescript"
 	],
 	"dependencies": {
-		"@typescript-eslint/eslint-plugin": "^4.9.1",
-		"@typescript-eslint/parser": "^4.9.1",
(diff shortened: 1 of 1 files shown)


sindresorhus:
> Globby is 2

Intentionally. I was waiting for fast-glob. https://github.com/mrmlnc/fast-glob/issues/240

fregante:
I was wondering why it wasn’t updated, I assumed it just needed some work due to breaking changes. 

What do you want to do with this PR then? Merge as is?

sindresorhus:
Is there something you need in these dependencies? Because I already did a "major" release 20 days ago, so not sure we should do it so soon again. I'm trying to not release too often.

fregante:
> Is there something you need in these dependencies?

Not really, it was supposed to be a "let's update everything and see what breaks" PR

PR #514, opened 2021-01-05
Clarify usage on Github Actions

codetheweb:
I wasted a bit of time getting annotations on Actions to work today; I think this is important enough to include in the documentation.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -303,6 +303,8 @@ See [eslint-config-xo-flow#use-with-xo](https://github.com/xojs/eslint-config-xo
 
 XO uses a different formatter when running in a GitHub Actions workflow to be able to get [inline annotations](https://developer.github.com/changes/2019-09-06-more-check-annotations-shown-in-files-changed-tab/). XO also disables warnings here.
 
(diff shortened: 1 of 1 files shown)


papb:
Hello! I thought it was not even possible to run `xo` before `setup-node`, since `xo` is installed via `npm` which in turn is installed when `node` is installed (with `setup-node`). Can you clarify what you were doing before?

codetheweb:
You don't have to use the `setup-node` action since Node + NPM are built into the base images.

Specific example: I'm using `ubuntu-latest`. Here's a run without `setup-node`:
[`f96e808` (#1)](https://github.com/Michigan-Tech-Courses/backend/pull/1/commits/f96e8083ec336421c67e1b71f6a3383462620c59)
and here's a run with `setup-node`:
[`ae5f730` (#1)](https://github.com/Michigan-Tech-Courses/backend/pull/1/commits/ae5f7302a52657ffd1c5b506ce6561e8132a406f)

sindresorhus:
This is exactly why I preferred Travis. There was no way to set up Node.js incorrectly.

PR #526, opened 2021-03-15
[readme] add to FAQ re: `--extend`

hangtwenty:
`xo` automatically detects the config files, by the names mentioned in the README. The `--extend` argument is not just an explicit version of that, it's specifically for extending eslint configs.

(Maybe it's obvious to some people but I think it is not obvious to newcomers 👋)


hangtwenty:
My mistake. I'm just having errors with my config and I'm not sure why.

PR #529, opened 2021-04-06
Add `--print-config` flag

andersk:
I often find myself needing this option to figure out which rules XO is using. If I learn about a new ESLint rule, I want to be able to check whether XO already enables it or something similar; if I upgrade XO, I want to be able to check whether it disabled any rules I was relying on; I want to be able to check that I configured `overrides` correctly and that it’s applied in the order I expect; etc.

This was largely based on #312 by @teehemkay.

### DIFF
--- a/cli-main.js
+++ b/cli-main.js
@@ -164,7 +164,7 @@ if (options.nodeVersion) {
 	if (options.nodeVersion === 'false') {
 		options.nodeVersion = false;
 	} else if (!semver.validRange(options.nodeVersion)) {
-		console.error('The `node-engine` option must be a valid semver range (for example `>=6`)');
+		console.error('The `--node-engine` flag must be a valid semver range (for example `>=6`)');
 		process.exit(1);
 	}
 }
@@ -185,7 +185,7 @@ if (options.nodeVersion) {
 		}
 
 		if (options.open) {
-			console.error('The `open` option is not supported on stdin');
+			console.error('The `--open` flag is not supported on stdin');
 			process.exit(1);
 		}
 
--- a/cli-main.js
+++ b/cli-main.js
@@ -30,6 +30,7 @@ const cli = meow(`
 	  --cwd=<dir>       Working directory for files
 	  --stdin           Validate/fix code from stdin
 	  --stdin-filename  Specify a filename for the --stdin option
+	  --print-config    Print the effective ESLint config for the given file
 
 	Examples
 	  $ xo
@@ -40,6 +41,7 @@ const cli = meow(`
 	  $ xo --plugin=react
 	  $ xo --plugin=html --extension=html
 	  $ echo 'const x=true' | xo --stdin --fix
+	  $ xo --print-config=index.js
 
 	Tips
 	  - Add XO to your project with \`npm init xo\`.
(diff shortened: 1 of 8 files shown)


cli-main.js
```diff
@@ -31,6 +31,7 @@ const cli = meow(`
 	  --cwd=<dir>       Working directory for files
 	  --stdin           Validate/fix code from stdin
 	  --stdin-filename  Specify a filename for the --stdin option
+	  --print-config    Print the ESLint configuration for the given file
```
sindresorhus, line 34:
```suggestion
	  --print-config    Print the effective ESLint config for the given file
```

Alternatively, "effective" => "resolved"

cli-main.js
```diff
@@ -174,7 +179,21 @@ if (options.nodeVersion) {
 }
 
 (async () => {
-	if (options.stdin) {
+	if (options.printConfig) {
+		if (input.length > 0) {
+			console.error('The `print-config` option must be used with exactly one file name');
```
sindresorhus, line 184:
```suggestion
			console.error('The `--print-config` flag must be used with exactly one filename');
```

sindresorhus: This message is confusing as it makes it sound like the user passed multiple filenames to `--print-config`.

cli-main.js
```diff
@@ -174,7 +179,21 @@ if (options.nodeVersion) {
...
+		}
+
+		if (options.stdin) {
+			console.error('The `print-config` option is not supported on stdin');
+			process.exit(1);
+		}
+
+		options.filename = options.printConfig;
+		const config = xo.getConfig(options);
+		console.log(JSON.stringify(config, null, '\t'));
```
sindresorhus, line 195:
```suggestion
		console.log(JSON.stringify(config, undefined, '\t'));
```

test/cli-main.js
```diff
@@ -167,3 +167,21 @@ test('extension option', async t => {
 	t.is(reports.length, 1);
 	t.true(reports[0].filePath.endsWith('.unknown'));
 });
+
+test('invalid print-config option with stdin', async t => {
+	const msg = 'The `print-config` option is not supported on stdin';
+	try {
+		await main(['--print-config', 'x.js', '--stdin'], {input: 'console.log()\n'});
```
sindresorhus, line 174: Use `t.throwsAsync` assertion.

test/print-config.js
```diff
@@ -0,0 +1,26 @@
+import path from 'path';
+import test from 'ava';
+import execa from 'execa';
+import tempWrite from 'temp-write';
+import fn from '..';
```
sindresorhus, line 5: Use a proper name.

sindresorhus (changes requested):

cli-main.js
andersk: That’s exactly the mistake this is intended to catch: `xo --print-config x.js y.js`. (This is parsed as `{flags: {printConfig: 'x.js'}, input: ['y.js']}`, but a user who writes that probably doesn’t know that.)

andersk (commented):
Applied all changes and squashed them into the `--print-config` commit. I also added three new commits to consistently apply these changes elsewhere in the existing code.

sindresorhus:
Looks good. Thank you 👍

PR #561, opened 2021-06-15
Support `xo.config.cjs` and `.xo-config.cjs`

### LINKED ISSUE #468: Node 14 ES Module: Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: xo.config.js
It seems impossible to load a JavaScript configuration when package.json contains `"type": "module"`.

In my project, there are complex conditions that necessitates using a JavaScript (not JSON) configuration, `xo.config.js`.
Recently I'm upgrading my project to [ES Modules](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html) by adding `"type": "module"` in package.json.

However, `xo` attempts to load `xo.config.js` using `require()`, and Node.js thinks it's a module that can only be loaded with `import()`, thus throwing error:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/user/project/xo.config.js
require() of ES modules is not supported.
require() of /home/user/project/xo.config.js from /home/user/project/node_modules/cosmiconfig/dist/loaders.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename xo.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/user/project/package.json.
```
It does not matter whether `xo.config.js` is written as ES Module or CommonJS.

Since removing `"type": "module"` is not an acceptable option, I tried renaming to `xo.config.cjs`, but `xo` does not load the configuration and is using its default configuration.

**package.json**
```json
{
  "private": true,
  "type": "module",
  "scripts": {
    "lint": "xo --fix"
  },
  "devDependencies": {
    "xo": "^0.30.0"
  }
}
```

**xo.config.js** or **xo.config.cjs**
```js
module.exports = {
	rules: {
		'object-curly-spacing': ['error', 'always']
	}
};
```

joelpurra:
- Fixes `ERR_REQUIRE_ESM` error from `cosmiconfig`.
- Enables using `xo.config.cjs`/`.xo-config.cjs` in ESM packages with `"type": "module"`.

Fixes #468.

See

- https://github.com/xojs/xo/issues/468
- https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

Sample error output; problem fixed by this commit.

```text
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /my-code/.xo-config.js
require() of ES modules is not supported.
require() of /my-code/.xo-config.js from /my-code/node_modules/cosmiconfig/dist/loaders.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename .xo-config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /my-code/package.json.
```

---

It's a start, but I don't think this fix is a golden bullet for ESM modules depending on `xo`. It fixes small-ish test cases without issue, but mixing ESM/non-ESM code may very well lead to further issues.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -120,7 +120,9 @@ const CONFIG_FILES = [
 	`.${MODULE_NAME}-config`,
 	`.${MODULE_NAME}-config.json`,
 	`.${MODULE_NAME}-config.js`,
-	`${MODULE_NAME}.config.js`
+	`.${MODULE_NAME}-config.cjs`,
+	`${MODULE_NAME}.config.js`,
+	`${MODULE_NAME}.config.cjs`
 ];
 
 const TSCONFIG_DEFFAULTS = {
--- a/readme.md
+++ b/readme.md
@@ -160,6 +160,14 @@ module.exports = {
 };
 ```
 
+4. For [ECMAScript module (ESM)](https://nodejs.org/api/esm.html) packages with [`"type": "module"`](https://nodejs.org/api/packages.html#packages_type), as a JavaScript module in `.xo-config.cjs` or `xo.config.cjs`:
+
+```js
+module.exports = {
+	space: true
+};
+```
+
 [Globals](https://eslint.org/docs/user-guide/configuring#specifying-globals) and [rules](https://eslint.org/docs/user-guide/configuring#configuring-rules) can be configured inline in files.
 
 ### envs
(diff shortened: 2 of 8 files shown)


sindresorhus:
You need to update https://github.com/xojs/xo#config

joelpurra:
@sindresorhus: yup. Mentioned the `.cjs` config files.

sindresorhus:
Thanks :)

PR #566, opened 2021-07-01
Convert project to module

fisker:
Questions:

- Do we still need support global `xo` to run local one?
- Don't know how to make `test/open-report.js` test pass, now skipped.


sindresorhus:
> Do we still need support global xo to run local one?

No, we should move to `npx xo`. Global XO doesn't currently work anyway: https://github.com/xojs/xo/issues/546

lib/options-manager.js
```diff
@@ -52,8 +46,7 @@ resolveFrom.silent = (moduleId, fromDirectory) => {
 	} catch { }
 };
 
-// TODO: Use `resolveModule(normalizePackageName(name), import.meta.url);` when moving to ESM then to `import.meta.resolve(normalizePackageName(name))` when supported
```
sindresorhus, line 55: Please preserve part of this todo comment.

sindresorhus (changes requested):

PR #582, opened 2021-08-01
Make cache directory be relative to cwd

### LINKED ISSUE #580: [bug] XO TS config parserOptions.project does not allow xo to be ran outside of dir
related to: https://github.com/Spence-S/vscode-linter-xo/issues/5 

Basically, xo uses the parserOption.project for use with typescript and it auto generates the tsconfig. However, it generates the tsconfig relative to where xo is called, and doesn't allow it to search from cwd argument.

This is all based off of a `cacheLocation` variable found here: https://github.com/Spence-S/xo/blob/f038f9c3b78430be829250b9438533aa5355bd74/lib/options-manager.js#L52

If we make the `cacheLocation` variable a function that takes a cwd argument, we can remedy this with no changes in behaviour while allowing both the cli and the api to be ran correctly on TS files outside of the current directory, which is a must for an xo editor extension for any editor.

**TL;DR**: xo needs to cache tsconfig relative to cwd option to work successfully in editor extensions.

Spence-S:
closes #580

### DIFF
--- a/test/options-manager.js
+++ b/test/options-manager.js
@@ -554,6 +554,8 @@ test('mergeWithFileConfig: typescript files', async t => {
 		semicolon: false,
 		ts: true,
 	};
+	const expectedConfigPath = new RegExp(`${slash(cwd)}/node_modules/.cache/xo-linter/tsconfig\\..*\\.json[\\/]?$`, 'u');
+	t.regex(slash(options.tsConfigPath), expectedConfigPath);
 	t.deepEqual(omit(options, 'tsConfigPath'), expected);
 	t.deepEqual(await readJson(options.tsConfigPath), {
 		extends: path.resolve(cwd, 'tsconfig.json'),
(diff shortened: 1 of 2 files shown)


test/options-manager.js
```diff
@@ -554,6 +554,9 @@ test('mergeWithFileConfig: typescript files', async t => {
 		semicolon: false,
 		ts: true,
 	};
+	// eslint-disable-next-line no-useless-escape
```
sindresorhus, line 557: Why are you disabling the lint rule? Is the rule incorrect?

.gitignore
```diff
@@ -4,3 +4,4 @@ yarn.lock
 test/fixtures/project/node_modules/.cache
 .nyc_output
 coverage
+.vscode
```
sindresorhus, line 7: This should be in your own global gitignore, not here.

PR #583, opened 2021-08-06
Simplify `lintFiles`

fisker:
This PR simplifies the `lintFiles` a lot, also changes how config is resolved, but tests are passing, and I ran it over some codebases, it seems works fine.

This also fixes a problem described [here](https://github.com/xojs/xo/issues/510#issuecomment-883020358)

There are some follow-up work can do, like DRY, use async api, I can send a separate PR after this get merged.


sindresorhus:
Thanks for working on this 🙌

sindresorhus:
> There are some follow-up work can do, like DRY, use async api, I can send a separate PR after this get merged.

That would be great 👍

PR #597, opened 2021-08-22
Remove `fs-extra` dependency

1000ch:
The usage of `fs-extra` is only related to reading/writing JSON currently, then they can be replaced with some combinations of native modules to reduce the dependency.

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -1,7 +1,7 @@
+import {promises as fs} from 'node:fs';
 import process from 'node:process';
 import os from 'node:os';
 import path from 'node:path';
-import fsExtra from 'fs-extra';
 import arrify from 'arrify';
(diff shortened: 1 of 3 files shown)


sindresorhus:
Tests are failing

1000ch:
> Tests are failing

Umm, cannot reproduce in my local environment... 👀

sindresorhus:
Thanks :)

devinrhode2:
Can you guys elaborate on the benefit of reducing dependencies? I guess install is slightly faster.. and yeah "fewer dependencies" makes it more robust

I wish node returned file contents as utf8 strings by default... maybe we should have a `readFile` util that's basically just: `fs.readFile(path, 'utf8')`? i.e. it has the `'utf8'` arg set as default, nothing else?

I'm working on some scripts of my own and just curious what you guys think is "good" or "best" practice here

PR #600, opened 2021-09-12
Drop some moot dependencies

1000ch:
Remove some replaceable/unused dependencies.

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -1,10 +1,9 @@
-import {promises as fs} from 'node:fs';
+import {existsSync, promises as fs} from 'node:fs';
 import process from 'node:process';
 import os from 'node:os';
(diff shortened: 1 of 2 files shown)


lib/options-manager.js
```diff
@@ -13,7 +12,6 @@ import {cosmiconfig, defaultLoaders} from 'cosmiconfig';
 import micromatch from 'micromatch';
 import JSON5 from 'json5';
 import toAbsoluteGlob from 'to-absolute-glob';
-import stringify from 'json-stable-stringify-without-jsonify';
```
sindresorhus, line 16: There's a reason this one is used.

1000ch: I see, just dropped the commit to remain this dependency.

PR #606, opened 2021-09-27
Fix `nodeVersion:false` not working

### LINKED ISSUE #456: nodeVersion is not respected in options
node: node v12.16.1
xojs: v0.28.3
example: https://github.com/ExiaSR/test-xoji-node-version

According to the [docs](https://github.com/xojs/xo#nodeversion), `nodeVersion` should work the same as `--node-version` cli argument.

However, running `yarn xo` and `yarn xo --node-version ">=8.0.0"` produced inconsistent result, where the former does not catch the `node/no-unsupported-features/es-syntax` error.

`yarn xo`

```bash
yarn run v1.22.4
warning package.json: No license field
$ /Users/michael/Code/playground/xojs-test/node_modules/.bin/xo

  test.js:4:7
  ✖  4:7  object is assigned a value but never used.  no-unused-vars

  1 error
error Command failed with exit code 1.
```

`yarn xo --node-version ">=8.0.0"`

```bash
yarn run v1.22.4
warning package.json: No license field
$ /Users/michael/Code/playground/xojs-test/node_modules/.bin/xo --node-version '>=8.0.0'

  test.js:4:7
  ✖  4:7   object is assigned a value but never used.                                                              no-unused-vars
  ✖  4:17  Rest/spread properties are not supported until Node.js 8.3.0. The configured version range is >=8.0.0.  node/no-unsupported-features/es-syntax
  ✖  4:32  Rest/spread properties are not supported until Node.js 8.3.0. The configured version range is >=8.0.0.  node/no-unsupported-features/es-syntax

  3 errors
error Command failed with exit code 1.
```

### LINKED ISSUE #598: `nodeVersion: false` doesn't disable the related rules
```
C:\Users\xmr\Desktop\foo>node -v && npm -v && ver
v14.17.5
6.14.14
Microsoft Windows [Version 10.0.19043.1165]

C:\Users\xmr\Desktop\foo>npm ls xo
foo@1.0.0 C:\Users\xmr\Desktop\foo
`-- xo@0.44.0
```

I was trying to disable the following two rules:

* "node/no-unsupported-features/es-builtins"
* "node/no-unsupported-features/es-syntax"

but it doesn't have any effect. So then I saw that there's the `nodeVersion` option that could control this. I tried setting it to `false` but it doesn't work either.

A quick patch would be to change https://github.com/xojs/xo/blob/bc6e05de5ad89e2e9d72070c7d4a1b04c0f4bd12/lib/options-manager.js#L205

```diff
-		...(enginesOptions && enginesOptions.node && semver.validRange(enginesOptions.node) ? {nodeVersion: enginesOptions.node} : {}),
+		...(xoOptions.nodeVersion !== false && enginesOptions && enginesOptions.node && semver.validRange(enginesOptions.node) ? {nodeVersion: enginesOptions.node} : {}),
```

But shouldn't it be possible to override the related rules?

XhmikosR:
Fixes #598

PS. I'm not familiar with the codebase and its tests, but if someone guides me I could try adding a test. Alternatively, feel free to push to the branch :)

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -207,6 +207,7 @@ const mergeOptions = (options, xoOptions = {}, enginesOptions = {}) => {
 		...options,
 	});
 
+	mergedOptions.nodeVersion = xoOptions.nodeVersion === false ? false : xoOptions.nodeVersion;
 	mergedOptions.extensions = [...DEFAULT_EXTENSION, ...(mergedOptions.extensions || [])];
 	mergedOptions.ignores = getIgnores(mergedOptions);
 


sindresorhus:
You could probably just update the test here: https://github.com/xojs/xo/blob/a2eada4aa1671c8de2ed2b157929c7d59bb9155f/test/options-manager.js#L233-L244

XhmikosR:
I added 3 tests like in `buildConfig: engines: undefined` but these were passing before anyway, so I'm not sure if they are right...

XhmikosR:
Unrelated to this but I noticed it while running the tests: there's something wrong with ava.

My VM has 4 threads yet there are spawned too many processes and makes the system unresponsive...

sindresorhus:
Did you manually verify that this PR fixes the issue for you?

XhmikosR:
When I tried it yeah, but now I see it's wrong. I'll mark it as a draft and have another try at it later.

XhmikosR:
This is what works for me, but tests break miserably. I'm going to close this for someone else more familiar with the codebase to handle.

Issue #614, opened 2021-10-06
Using prettier, but seeing `import {Foo}` (object-curly-spacing)


devinrhode2:
I have `prettier: true` in my xo config, but, I'm seeing this formatting:
```
import {Foo} from './foo'
```
This is fundamentally the "object-curly-spacing" style rule from eslint (or `@typescript-eslint/object-curly-spacing`).

I see this is set inside eslint-config-xo and eslint-config-xo-typescript, to `["error", "never"]` - "never" meaning we never want spacing around `Foo`.

I confirmed this by setting `prettier: false` inside my xo config, and inside vscode, I see a warning regarding this style rule.
![Screen Shot 2021-10-05 at 08 59 44 PM](https://user-images.githubusercontent.com/539816/136132751-01197562-6dfb-4bac-8cec-9865d1e9a5b6.png)

As a sanity check, I did run prettier directly:
```
yarn prettier --write /full/path/to/file.ts
```
And it restores spaces around `Foo`, resulting in:
```
import { Foo } from './foo'
```

I tried running `yarn prettier --write /full/path/to/file.ts` with multiple different prettier versions, 2.2.1, 2.3.2, 2.4.1, all of them restore spaces around `Foo`.

devinrhode2:
Q: Does the vscode extension supports codeActionsOnSave?
=====
My general thinking is that we should address this at an architectural level, so that any formatting inconsistencies are not possible. I think that prettier should run after eslint, and eslint should ignore all formatting issues.

However, I would like to avoid writing eslint fixed code and then telling prettier to read from disk. Passing an ast directly to prettier would be awesome, but would be quite the undertaking.

devinrhode2:
@fisker I noticed you have some PRs for eslint-plugin-prettier, which fix the arrow-body-style issues

Any thoughts here? My interest is in improving performance, but main issue here is the inconsistent formatting

Spence-S:
@devinrhode2 xo is only a wrapper around eslint - prettier is only ran via eslint-plugin-prettier from xo. You can configure prettier by adding a prettier field to your package.json or a .prettierrc file, you probably get this. Yes, I don't think xo follows the same defaults as prettier itself, but xo's default prettier config does NOT conflict with the xo eslint rules. - it is kind of confusing. 

Best idea is to let xo manage both formatting and linting for you completely. Or turn prettier off in xo and configure prettier to 

[...]

PR #616, opened 2021-10-06
Clarify prettier defaulting behavior

devinrhode2:
This definitely sent me down a rabbit hole: #614

TODO:
1. Cleanup source for `bracketSameLine`. XO Does not need to set it to `false`, since that is prettier's current default. (I'm assuming it was different in the past).
2. Verify this actually works, check when errors are actually thrown

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -221,14 +221,23 @@ Default: `false`
 
 Format code with [Prettier](https://github.com/prettier/prettier).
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
> Cleanup source for bracketSameLine. XO Does not need to set it to false, since that is prettier's current default. (I'm assuming it was different in the past).

PR welcome

devinrhode2:
I would argue that XO should not be changing these defaults, but it's probably not worth the breaking change for existing consumers. Maybe there's a way to detect an upgrade and then just modify the users prettier config, or just print a warning.

sindresorhus:
Changing the defaults is very much intentional.

PR #617, opened 2021-10-08
Revert fisker's memory regression

devinrhode2:
I am happy with the git work, but still need to do some manual sanity checking, and probably run test suite

@fisker you are totally welcome to change anything here. For example, "use `xo` to lint `xo` was lost as collateral damage, but we can restore that. Likewise I think the "ignores" fix you had in the same commit could also be restored.

Here, I was just trying to do the cleanest possible revert.

This could be considered like a v0.45.1 but since it's also reverting fisker's fixes, it could also be considered like a v0.43.1. It's a mutant. Probably 0.43.1 would make the most sense. (If review goes well, and sindre wants to publish it, of course)

Closes #599 but also re-opens #583

PR #619, opened 2021-10-15
Revert memory regression #617 (and revert "remove fs-extra")

devinrhode2:
taking the truly brain-dead approach and also reverting the "remove fs-extra" commit, fingers crossed build passes

I'm not concerned with seeing this PR merged
- primarily, I am just curious if the build will pass
- secondarily, others can see this PR, if they are suffering from memory regression, this branch may help them


devinrhode2:
I'm not sure how much better memory usage is, as I'm always running XO with 16gb `"xo": "/usr/bin/time node --max-old-space-size=16384 --stack-trace-limit=25 ./node_modules/.bin/xo",`, but the main regression was solved (see original issue)

PR #620, opened 2021-10-16
Rename allowList/denyList in `prevent-abbreviations`

fregante:
Arrays don't necessarily need to be named `list`. These lists are can just be named allow/allowed/include or exclude/except/exclusions. `deny` itself feels kinda awkward too because it usually applies to "access" only, not other kinds of generic exclusions lists.

### DIFF
--- a/config/plugins.cjs
+++ b/config/plugins.cjs
@@ -44,7 +44,7 @@ module.exports = {
 				replacements: {
 					// https://thenextweb.com/dd/2020/07/13/linux-kernel-will-no-longer-use-terms-blacklist-and-slave/
 					whitelist: {
-						allow: true,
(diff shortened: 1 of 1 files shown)


fregante:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/942#issuecomment-751315142 😃 

Every case has its own solution, but we usually don't append `list` to array names.

sindresorhus:
`inclusions` and `exclusions`, to be coherent?

sindresorhus:
@fregante Intentional that you didn't do `inclusions`?

fregante:
Yes, I think `include/exclude` is more common/readable

PR #632, opened 2021-11-01
Fix handling of tsconfig's

### LINKED ISSUE #629: "parserOptions.project" errors occur when users tsconfig does not include a file
If a user has a TS config and some JS/TS files are not included - an error will be produced for that file instead of proper linting results. 

Regression Introduced by #622 as noted in #627

Spence-S:
- fixes #629 

- will only use default tsconfig if no other can be found.

- Now detects if tsconfig properly applies to file. If not - creates a temp tsconfig that will cover the file in question which extends any found tsconfig (this is the same as it was) and will prevent any `parserOptions.project` errors. (this is very similar to what happened prior to #629 regression)

- Will not create a new file if not needed.  

- tsconfig hashes are more predictable now and won't create different hashes for different node versions, which could previously cause confusion.


sindresorhus:
Seems to fail on Node.js 16 in this TS project: https://github.com/sindresorhus/ow/runs/4091470790?check_suite_focus=true

https://github.com/sindresorhus/ow/pull/225/commits/db0eef899a8f2a26aef84e5381a3af568947df88

Spence-S:
@sindresorhus - I don't think that failure is related to this PR - its a module resolution error. Nothing in this PR changed any module resolution stuff. That failure was from https://github.com/xojs/xo/blob/4c51f2341fed6d767486bf0a8c4c354d5ff5532e/lib/options-manager.js#L479 - could this be related to #624 ?

sindresorhus:
Thanks for fixing this 🙌

PR #639, opened 2021-11-25
Enforce extensions on TypeScript files as well

fregante:
I think this config should be dropped. Reasons:

- it actually "breaks" the rule for .ts/tsx files because it allows them to be imported without extension

	1. create `file.ts`
	2. `import './file'` in `another.ts`
	3. No errors! 😰 

- it duplicates TypeScript’s own error, but with a worse message:

	Screen Shot 3

### DIFF
--- a/config/plugins.cjs
+++ b/config/plugins.cjs
@@ -193,12 +193,7 @@ module.exports = {
 			'error',
 			'always',
 			{
-				ignorePackages: true,
-				// TypeScript doesn't yet support using extensions and fails with error TS2691.
-				pattern: {
-					ts: 'never',
-					tsx: 'never',
-				},
+				ignorePackages: true
 			},
 		],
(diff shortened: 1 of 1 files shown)


fregante:
Also it appears that this configuration actually _breaks_ the rule.

Try:

- create `file.ts`
- `import './file'`

The rule does not complain because `.ts` is among the forbidden extensions, so that import is allowed extension-less.

fregante:
The side effect of this change is that the rule will now suggest to use `.ts` extension in that case, but 

```
  2:27  error  Missing file extension "ts" for "./sender"  import/extensions
```

but once you add it, TypeScript will ask you to change it to .js, which is fine and is really TypeScript’s limit and rule here.

```
error TS2691: An import path cannot end with a '.ts' extension. Consider importing './sender.js' instead.
```

PR #665, opened 2022-05-14
Add `--fix-dry-run` option

### LINKED ISSUE #663: Support `--fix-dry-run` flag
I think it would be great if `xo` could format the source file in vim automatically.

https://github.com/jose-elias-alvarez/null-ls.nvim/pull/707#issuecomment-1050118227

In the above vim plugin's PR, `xo` was added as a linter, but had not been added as a formatter since `fix` flag automatically save the changes to the file system.

It seems supporting `--fix-dry-run` flag might help these libraries to support `xo` as formatter.

Many thanks to this awesome library.

Ref:
https://eslint.org/docs/user-guide/command-line-interface#options

jopemachine:
Fixes #663.

Eslint doesn't expose `fixDryRun` option in programmatic API. 
`fixDryRun` is [internally translated to `fix` option](https://github.com/eslint/eslint/blob/main/lib/cli.js#L96) and don't call `ESLint.outputFixes`.
(Ref: https://github.com/eslint/eslint/discussions/15877)

So I think `xo` could handle this option similarly.
`fixDryRun` is translated to `fix` option before passing to ESLint constructor, and `xo.outputFixes` would not be called since 
the `fix` option is false.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -13,6 +13,7 @@ const cli = meow(`
 
 	Options
 	  --fix             Automagically fix issues
+	  --fix-dry-run     Automatically fix problems without saving the changes to the file system
 	  --reporter        Reporter to use
 	  --env             Environment preset  [Can be set multiple times]
 	  --global          Global variable  [Can be set multiple times]
@@ -51,6 +52,9 @@ const cli = meow(`
 	booleanDefault: undefined,
 	flags: {
 		fix: {
+			type: 'boolean',
+		},
+		fixDryRun: {
 			type: 'boolean',
 		},
 		reporter: {
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -309,6 +309,10 @@ const buildESLintConfig = options => config => {
 		};
 	}
 
+	// eslint doesn't expose `fixDryRun` option in programmatic API.
+	// Instead of using `fixDryRun` itself, the option is internally translated to `fix` option in eslint's CLI module.
+	// (Ref: https://github.com/eslint/eslint/blob/901ce0f1e32ea1e9e10ce4d8b37c0d750007a3c5/lib/cli.js#L96)
+	// xo also could handle this in the same way.
 	if (options.fixDryRun) {
 		config.fix = true;
 	}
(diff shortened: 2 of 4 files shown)


lib/options-manager.js
```diff
@@ -309,6 +309,10 @@ const buildESLintConfig = options => config => {
 		};
 	}
 
+	if (options.fixDryRun) {
```
sindresorhus, line 312: I think this could use a code comment on how it works. It's not immediate obvious how it works since you only see here that it sets `.fix = true`.

cli.js
```diff
@@ -13,6 +13,7 @@ const cli = meow(`
 
 	Options
 	  --fix             Automagically fix issues
+	  --fix-dry-run     Automatically fix problems without saving the changes to the file system
```
sindresorhus, line 16: I feel like it's unclear to the user how this would even work. The user will ask themselves: "If it does not save to the file system, what's the point of it?"

lib/options-manager.js
```diff
@@ -309,6 +309,13 @@ const buildESLintConfig = options => config => {
 		};
 	}
 
+	// ESLint does not expose a `fixDryRun` option in the programmatic API.
+	// Instead of using `fixDryRun` itself, the option is internally translated to `fix` option in ESLint's CLI module.
+	// https://github.com/eslint/eslint/blob/901ce0f1e32ea1e9e10ce4d8b37c0d750007a3c5/lib/cli.js#L96
```
sindresorhus, line 314: This does not make it clear how it then manages to be a "dry-run". The important part is missing.

sindresorhus (changes requested):

cli.js
jopemachine: --fix-dry-run     Automatically fix problems without saving the changes to the file system.
                      Because the default formatter does not output the fixed code, you'll have to use another one (e.g. json)
                      This flag can be useful for integrations which need to autofix text from command line without saving it to the filesystem.

How about the above one?
I'm worried about only this option's description looks too verbose because the other options are all single lines.
These descriptions are copy-pasted from https://eslint.org/docs/user-guide/command-line-interface#--fix-dry-run

mvllow: Perhaps "Preview fixed file contents" or "Print fixed file contents"?

```suggestion
	  --fix-dry-run     Preview fixed file contents
```

PR #672, opened 2022-06-21
fix: honor user configured rules

### LINKED ISSUE #613: Cannot disable `eslint-plugin-node` and `eslint-plugin-n` rules
Adding `"node/no-unsupported-features/es-syntax": "off"` to the rules section do not disable all related errors.

This is particularly problematic for me since there's a false positive due to dynamic `import()` (issue: https://github.com/mysticatea/eslint-plugin-node/issues/250) and I cannot find a way to turn off the error.

Is there something I missed?

aladdin-add:
- [x] fix failing tests
- [ ] add a test

atm, users cannot disable the rules, as the defaults was always applied.
fixes https://github.com/xojs/xo/issues/670 fixes #613

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -327,9 +327,11 @@ const buildXOConfig = options => config => {
 	}
 
 	if (options.nodeVersion) {
-		config.baseConfig.rules['n/no-unsupported-features/es-builtins'] = ['error', {version: options.nodeVersion}];
-		config.baseConfig.rules['n/no-unsupported-features/es-syntax'] = ['error', {version: options.nodeVersion, ignores: ['modules']}];
-		config.baseConfig.rules['n/no-unsupported-features/node-builtins'] = ['error', {version: options.nodeVersion}];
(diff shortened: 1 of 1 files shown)


aladdin-add:
the build is failing as `?.` and `??` was not supported in node v12, will try to make a change later.

fregante:
Is this still mergeable or necessary? Node 12 is long gone

sindresorhus:
It may be moot with flat config and I also don't want to take the risk of breakage, since XO is going to change a lot anyway with flat config.

fregante:
> I also don't want to take the risk of breakage

Understandable, but due to both https://github.com/xojs/xo/issues/613 and https://github.com/xojs/xo/issues/598, I'm seeing errors related to Node in a non-Node project, and they require inline exclusions.

One example is:

```
✖    7:9   The navigator is still an experimental feature The configured version range is >= 20.                                                                          n/no-unsupported-features/node-builtins
```

PR #674, opened 2022-07-19
include rulesMeta by default for all linting

Spence-S:
Simple addition for use in the vscode extension. 

I had added a different package to get the rule docs, but it is broken for mono-repo based plugins, and this change was simpler than fixing the third party rule-docs package.

Note: we _may_ want to add a similar change to lintFiles as we currently construct our urls in eslint-formatter-pretty with the broken package I mentioned. 

related to: https://github.com/xojs/vscode-linter-xo/issues/112

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,7 +66,10 @@ const lintText = async (string, options) => {
 	}
 
 	const report = await eslint.lintText(string, {filePath, warnIgnored});
-	return processReport(report, {isQuiet});
+
+	const rulesMeta = eslint.getRulesMetaForResults(report);
+
+	return processReport(report, {isQuiet, rulesMeta});
 };
 
 const lintFiles = async (patterns, options) => {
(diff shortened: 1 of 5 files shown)


sindresorhus:
> Note: we may want to add a similar change to lintFiles as we currently construct our urls in eslint-formatter-pretty with the broken package I mentioned.

👍 Is this something you would be willing to do in this PR?

index.js
```diff
@@ -66,7 +66,14 @@ const lintText = async (string, options) => {
 	}
 
 	const report = await eslint.lintText(string, {filePath, warnIgnored});
-	return processReport(report, {isQuiet});
+
+	let rulesMetaForResults;
+
+	if (options && options.rulesMetaForResults === true) {
+		rulesMetaForResults = eslint.getRulesMetaForResults(report);
```
sindresorhus, line 73: Does this call have any noticeable overhead? If not, we could maybe just do it by default without an option.

Spence-S:
> 👍 Is this something you would be willing to do in this PR?

Sure! The change is simple but a little more involved than in lintText, will need to update a few tests as well. Give me a couple of days for the changes and I will ping you again for re-review.

index.js
sindresorhus: Then I suggest we just do it by default without an option.

Spence-S:
Per the latest update to this PR: 
- adds rulesMeta step for all linting and merges into the report
- properly passes rulesMeta to the formatter in the cli
- tests rules meta in lint-files and lint-text tests by modifying the `hasRule` function to check that corresponding metadata is there as well.

sindresorhus:
https://github.com/xojs/xo/releases/tag/v0.51.0

PR #677, opened 2022-08-07
feat: full tsconfig resolution (closes #637 closes #661)

### LINKED ISSUE #637: @typescript-eslint/parser error with TypeScript test files
I'd like to use XO to lint and format my TypeScript tests in addition to my source code.

When I add a test file to my project, I get the following error from XO:

```
  src/index.test.ts
  ✖  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: src/index.test.ts.
The file must be included in at least one of the projects provided.  

  1 error
```

I think this is happening because my test files are excluded in TSConfig.

Here are some minimal reproductions in StackBlitz:

- [XO + TS with `exclude`](https://stackblitz.com/edit/node-pm6apl?file=tsconfig.json)
    - Run `npm test` to see above error
- [XO + TS without `exclude`](https://stackblitz.com/edit/node-pxywud?file=tsconfig.json)
    - Run `npm test` to see no error

### LINKED ISSUE #661: File matching to determine generated tsconfig does not take relative directories into account
## Overview

xo attempts to check if a given file that it's linting matches a `tsconfig.json` passed into `parserOptions.project`. If the file being checked does not match the tsconfig (according to minimatch), a temporary tsconfig will be created.

https://github.com/xojs/xo/blob/352f20c0b0572c674f519cdf98a46fa0f35a96d9/lib/options-manager.js#L178-L181

However, `options.filePath` here is an absolute filepath, and the `includes` glob of a given `tsconfig.json` may be specified as a relative path from the config file itself. If the tsconfig in question is not in the current working directory, this check will erroneously fail.

I noticed this issue due to xojs/vscode-linter-xo#110. I thought I had given `xo` a good `parserOptions.project` path, but `xo` and the `vscode-linter-xo` started generating competing (and different!) temporary tsconfigs in the cache, eventually causing the editor extension to start choking on parser errors.

### Project structure
```
- my-project/
    - package.json 
    - config/
        - tsconfig.lint.json
    - src/
        - some-file.ts
```

### tsconfig.lint.json

```js
{
  ...
  "include": ["../src/**/*"]
}
```

### Behavior

`/path/to/my-project/src/some-file.ts` does not match the glob `../src/**/*`, so a temporary tsconfig is created in `node_modules/.cache/xo-linter`, even though `tsc` would recognize the config as matching that file.

```json
{
  "extends": "/path/to/my-project/config/tsconfig.lint.json",
  "files": ["/path/to/my-project/src/some-file.ts"],
  "include": [],
  "exclude": []
}
```

## Proposal

Before passing `options.filePath` to `micromatch`, I think it should be made relative to `options.tsConfigPath` if it exists. If this is an acceptable solution, I'm happy open a PR!

```js
micromatch.contains("/path/to/my-project/src/some-file.ts", ["src/**/*"])
> true
micromatch.contains("/path/to/my-project/src/some-file.ts", ["../src/**/*"])
> false
micromatch.contains("../src/some-file.ts", ["../src/**/*"])
> true
```

## Workaround

This issue is pretty straightforward to workaround by placing the tsconfig given to `parserOptions.project` in the project root rather than a subdirectory, so that a glob like `src/**/*` will match an absolute path.

Spence-S:
This PR causes xo to recursively build tsconfig for files by merging each extends block into the final tsconfig. 

This allows us to have the proper information to correctly determine typescript options/locations and get rid of the bugs as its now much more similar to how the actual tsconfig algorithm works.

It also adds some more complex tsconfig related tests.

Fixes #637 
Fixes #661


lib/options-manager.js
```diff
@@ -607,6 +635,58 @@ const getOptionGroups = async (files, options) => {
...
+
+	return recursiveBuildTsConfig(tsConfig, basePath);
+}
+
+// Convert all include, files, and exclude to absolute paths
+// and or globs. This works because ts only allows simple glob subset
+const tsConfigAbsPaths = (tsConfig, tsConfigPath) => {
+	if (Array.isArray(tsConfig.files)) {
+		tsConfig.files = tsConfig.files.map(
+			f => path.isAbsolute(f) ? f : path.resolve(path.dirname(tsConfigPath), f),
```
sindresorhus, line 671: `path.dirname(tsConfigPath)` can be done once at the top-level in the function.

lib/options-manager.js
```diff
@@ -607,6 +635,58 @@ const getOptionGroups = async (files, options) => {
...
+		);
+	}
+
+	if (Array.isArray(tsConfig.include)) {
+		tsConfig.include = tsConfig.include.map(
+			f => path.isAbsolute(f) ? f : path.resolve(path.dirname(tsConfigPath), f),
```
sindresorhus, line 677: Don't use one-character variable name.

lib/options-manager.js
```diff
@@ -607,6 +635,58 @@ const getOptionGroups = async (files, options) => {
...
+		);
+	}
+
+	if (Array.isArray(tsConfig.exclude)) {
+		tsConfig.exclude = tsConfig.exclude.map(
+			f => path.isAbsolute(f) ? f : path.resolve(path.dirname(tsConfigPath), f),
```
sindresorhus, line 683: The `path.isAbsolute` check is moot. `path.resolve` already passes `f` through if absolute.

lib/options-manager.js
```diff
@@ -607,6 +635,58 @@ const getOptionGroups = async (files, options) => {
...
+		...baseTsConfig,
+		...tsConfig,
+	};
+
+	return recursiveBuildTsConfig(tsConfig, basePath);
+}
+
+// Convert all include, files, and exclude to absolute paths
+// and or globs. This works because ts only allows simple glob subset
+const tsConfigAbsPaths = (tsConfig, tsConfigPath) => {
```
sindresorhus, line 668:
```suggestion
const tsConfigResolvePaths = (tsConfig, tsConfigPath) => {
```

sindresorhus (changes requested):

PR #680, opened 2022-08-25
fix: ensure tsconfig lookups and parsing work as expected

### LINKED ISSUE #679: 0.52 extending tsconfig from node_modules causes error
I assume this is from #677 

```plaintext
Error: ENOENT: no such file or directory, open '<path>/<project-working-directory>/<extends-string>'
```

The extends string seems to be exactly what is specified in the `tsconfig.json` `extends` value.

I tried both `@sindresorhus/tsconfig` and `@sindresorhus/tsconfig/tsconfig.json` and both fail.
This happens both locally on macOS and Linux and in GitHub CI Runners.

Spence-S:
- changes tsconfig lookups to use `require.resolve` instead of building the path relative to the cwd
- use JSON5 to parse tsconfigs properly

closes #679


sindresorhus:
Thanks for the quick fix :)

PR #683, opened 2022-08-25
fix: tsconfig resolution quirks

### LINKED ISSUE #682: Error: Cannot find module '<tsconfig from node_modules>'
## Bug description

When I run `xo` I get this error:

```
Error: Cannot find module '@namespace/package'
Require stack:
- /Users/jonah/programming/xo-bug/tsconfig.json
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:956:15)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at recursiveBuildTsConfig (file:///Users/jonah/programming/xo-bug/node_modules/xo/lib/options-manager.js:649:27)
    at handleTSConfig (file:///Users/jonah/programming/xo-bug/node_modules/xo/lib/options-manager.js:180:28)
    at async mergeWithFileConfig (file:///Users/jonah/programming/xo-bug/node_modules/xo/lib/options-manager.js:139:13)
    at async parseOptions (file:///Users/jonah/programming/xo-bug/node_modules/xo/lib/options-manager.js:586:51)
    at async Promise.all (index 0)
    at async getOptionGroups (file:///Users/jonah/programming/xo-bug/node_modules/xo/lib/options-manager.js:602:21)
    at async Object.lintFiles (file:///Users/jonah/programming/xo-bug/node_modules/xo/index.js:78:17)
    at async file:///Users/jonah/programming/xo-bug/node_modules/xo/cli.js:211:18
```

This is my `tsconfig.json`:
```json
{
	"compilerOptions": {
		"declaration": true,
		"inlineSourceMap": true,
		"moduleResolution": "node",
		"outDir": "dist"
	},
	"extends": "@tsconfig/node16",
	"include": ["src"]
}
```

## Bug reproduction

Here's a fully reproducible project:
[xo-bug.zip](https://github.com/xojs/xo/files/9428015/xo-bug.zip)
1. Download & extract
2. `yarn`
3. `yarn test`
4. Error occurs

## Fix

I believe this is a bug introduced in #677.

The fix is probably replacing this:
https://github.com/xojs/xo/blob/c8ec522edc5418c583d647025889e8bcd18e091d/lib/options-manager.js#L642-L644

with something like this:
```js
 if (!tsConfig.extends || (typeof tsConfig.extends === 'string' && tsConfig.extends.includes('node_modules') || tsConfig.startsWith('@'))) { 
 	return tsConfig; 
 } 
```

Spence-S:
closes #682 

- fixes an issue where a valid tsconfig with comments could potentially error
- adds a failsafe for some quirky behavior around resolution for `extends`

Unfortunately tsconfig resolution is less straightforward that I thought, and apologies for the churn around this. Hopefully this change will now support the vast majority of ways people can use a tsconfig 

I have played around with using TS directly for parsing the tsconfig it works, but the downsides are that it breaks a lot of tests, is a relatively big refactoring, and it is all sync methods.


lib/options-manager.js
```diff
@@ -644,9 +644,22 @@ async function recursiveBuildTsConfig(tsConfig, tsConfigPath) {
...
+	try {
+		basePath = require.resolve(tsConfig.extends);
+	} catch (error) {
+		// Tsconfig resolution is odd, It allows behavior that is not exactly like node resolution
+		// therefore we attempt to smooth this out here with this hack
+		try {
+			basePath = require.resolve(path.join(tsConfig.extends, 'tsconfig.json'));
+		} catch {
+			// Throw the orginal resolution error to let the user know their extends block is invalid
+			throw error;
```
Spence-S, line 659: alternatively we could just swallow this error, and when that happens we would just generate the needed tsconfig for the file. That may be the better way to handle this.

sindresorhus: I don't think we should swallow errors like that. It sounds like a pain to debug that it's working, but with the incorrect config.

PR #686, opened 2022-08-28
Relative 'extends' no longer error out in the XO VS Code extension, nor if running XO from a different folder

andrewdbond:
Relative 'extends' no longer error out in the XO VS Code extension, nor if running XO from a different folder.

Fixes xojs/xo#685

Addressed issue by altering the existsSync call. By default [node:fs resolves paths relative to the directory returned by process.cwd](https://nodejs.org/api/fs.html#string-paths). For the XO VS Code extension, process.cwd() references the VS Code program folder.

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -413,7 +413,7 @@ const buildExtendsConfig = options => config => {
 	if (options.extends && options.extends.length > 0) {
 		const configs = options.extends.map(name => {
 			// Don't do anything if it's a filepath
-			if (existsSync(name)) {
+			if (existsSync(path.resolve(options.cwd || process.cwd(), name))) {
 				return name;
 			}
 


sindresorhus:
// @Spence-S

Spence-S:
@sindresorhus - this is not related to recent changes, I don't believe. If I am understanding right, this is about extending the configuration from a relative dir -- it's always resolved from `process.cwd()` which causes it to not be located correctly if the config dir does not match. It is not directly related to the extension. I think it is a valid bug.

@andrewdbond this is probably a good solution/fix but I would like to see a failing test or 2 added so this problem won't sneak back in later and we can clearly see the use case being solved.

andrewdbond:
Added to ./test/, thanks.
// @Spence-S

PR #687, opened 2022-08-30
Fix windows directory  separator

skarab42:
@sindresorhus Under windows some tests on file names did not match because they were compared with different directory separators. This PR fixes this issue using the `slash` lib where it was required for my case, it is possible that for other configurations (under windows) it should be done elsewhere as well.

Fixes https://github.com/sindresorhus/type-fest/issues/451

### DIFF
--- a/lib/options-manager.js
+++ b/lib/options-manager.js
@@ -16,6 +16,7 @@ import stringify from 'json-stable-stringify-without-jsonify';
 import {Legacy} from '@eslint/eslintrc';
 import createEsmUtils from 'esm-utils';
 import MurmurHash3 from 'imurmurhash';
+import slash from 'slash';
 import {
 	DEFAULT_IGNORES,
 	DEFAULT_EXTENSION,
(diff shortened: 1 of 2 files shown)


lib/options-manager.js
```diff
@@ -174,6 +175,10 @@ const handleTSConfig = async options => {
 		options.tsConfig = searchResults.config;
 	}
 
+	if (options.tsConfigPath) {
+		options.tsConfigPath = slash(options.tsConfigPath);
```
sindresorhus, line 179: I think the `slash` calls should only be done right before passing some paths to globbing, as it's the globbing that does not support backward slashes, everything else does.

skarab42: In fact it was not required, just needed to update the tests.

test/options-manager.js
```diff
@@ -645,7 +645,7 @@ test('mergeWithFileConfig: auto generated ts config extends found ts config if f
 test('mergeWithFileConfig: used found ts config if file is covered', async t => {
 	const cwd = path.resolve('fixtures', 'typescript', 'extends-config');
 	const filePath = path.resolve(cwd, 'foo.ts');
-	const expectedConfigPath = path.resolve(cwd, 'tsconfig.json');
+	const expectedConfigPath = slash(path.resolve(cwd, 'tsconfig.json'));
```
sindresorhus, line 648: In the other tests, you use `slash` on the `options.tsConfigPath`.

skarab42: Right, here no need for either.

PR #710, opened 2023-03-31
Bump typescript to 5.0.3

### LINKED ISSUE #709: Error with typescript 5.0 features and yarn
When using yarn and typescript 5.0 features (eg. verbatimModuleSyntax) it throws because xo uses `typescript@^4.9.3` and then doesn't understand the new options.  
Yarn won't bump it to 5.0 because the bump is a major change. This is an issue with both regular yarn and yarn berry (but not npm).  
I've bumped it by modifying `yarn.lock` and nothing broke, so it should probably be an easy fix

I've created a repo: https://github.com/melusc/xo-ts5.0-yarn

melusc:
Fixes #709

I can also update the other dependencies, if you want

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
 		"semver": "^7.3.8",
 		"slash": "^5.0.0",
 		"to-absolute-glob": "^2.0.2",
-		"typescript": "^4.9.3"
+		"typescript": "^5.0.3"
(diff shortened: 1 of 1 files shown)


melusc:
Guess it is more breaking than I anticipated (#711)

PR #712, opened 2023-04-03
Feat: support ts5

### LINKED ISSUE #709: Error with typescript 5.0 features and yarn

### LINKED ISSUE #711: Support for Typescript v5 `extends` array
Typescript v5 now [supports multiple base configurations](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#supporting-multiple-configuration-files-in-extends) through `extends` as an array.

Unfortunately `xo` throws an error as it expects to be a string:

```sh
❯ yarn lint
TypeError [ERR_INVALID_ARG_TYPE]: The "request" argument must be of type string. Received an instance of Array
    at new NodeError (node:internal/errors:399:5)
    at validateString (node:internal/validators:163:11)
    at Function.resolve (node:internal/modules/cjs/helpers:115:5)
    at recursiveBuildTsConfig (file:///Users/user/project/node_modules/xo/lib/options-manager.js:652:22)
    at handleTSConfig (file:///Users/user/project/node_modules/xo/lib/options-manager.js:181:28)
    at async mergeWithFileConfig (file:///Users/user/project/node_modules/xo/lib/options-manager.js:140:13)
    at async parseOptions (file:///Users/user/project/node_modules/xo/lib/options-manager.js:587:51)
    at async Promise.all (index 2)
    at async getOptionGroups (file:///Users/user/project/node_modules/xo/lib/options-manager.js:603:21)
    at async Object.lintFiles (file:///Users/user/project/node_modules/xo/index.js:78:17)
```

spence-s:
- bump ts dep, although not sure we even use it.
- add `get-tsconfig` to handle tsconfig extends resolution as its much more robust than the logic we had and supports ts5 already. 

closes #711 
closes #709


package.json
```diff
@@ -93,7 +93,8 @@
 		"semver": "^7.3.8",
 		"slash": "^5.0.0",
 		"to-absolute-glob": "^2.0.2",
-		"typescript": "^4.9.3"
+		"typescript": "^5.0.3",
+		"xo": "file:."
```
sindresorhus, line 97: I assume this was not intentional?

PR #715, opened 2023-04-25
Upgrade bundled typescript-eslint

andersk:
Among other improvements, this fixes a crash in the `@typescript-eslint/unified-signatures` rule: typescript-eslint/typescript-eslint#6940.

(Alternatively, since Node.js 14 reaches [end-of-life](https://github.com/nodejs/Release#release-schedule) this weekend, maybe we can revert #624 so that users can get the latest typescript-eslint independently again?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
 	],
 	"dependencies": {
 		"@eslint/eslintrc": "^1.3.3",
-		"@typescript-eslint/eslint-plugin": "^5.57.1",
-		"@typescript-eslint/parser": "^5.57.1",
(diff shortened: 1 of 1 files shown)


sindresorhus:
> (Alternatively, since Node.js 14 reaches [end-of-life](https://github.com/nodejs/Release#release-schedule) this weekend, maybe we can revert https://github.com/xojs/xo/pull/624 so that users can get the latest typescript-eslint independently again?)

Yes, that's the plan. But I don't want to require Node.js 16 just yet.

Issue #760, opened 2024-07-20
Cannot disable `n/no-unsupported-features/node-builtins`


melusc:
Setting `"n/no-unsupported-features/node-builtins": "off"` gets ignored. Probably also applies to all `n/no-unsupported-features/*`. 

See <https://github.com/melusc/xo-bug-n-no-supported-features> for a minimal reproducible example. You can see in <https://github.com/melusc/xo-bug-n-no-supported-features/blob/main/resulting-config.json> (generated with `--print-config`) that it is still enabled, and running `npx xo` still gives that error.

melusc:
duplicate of #598, sorry

PR #761, opened 2024-07-20
fix: Don't override rules from user's config

### LINKED ISSUE #598: `nodeVersion: false` doesn't disable the related rules

melusc:
If the user disables any of `n/no-unsupported-features/*` or otherwise overrides it in their config it will now keep that setting.

Fixes #598 and #760
Closes #672


melusc:
I haven't looked into it, but I would imagine <https://github.com/xojs/xo/blob/112e6579181d4080897c4ff1e2d1ab45da0180a3/lib/options-manager.js#L364> and similar lines should also use `??=` instead to respect the user's config. I can make a PR or include that in this PR if desired

sindresorhus:
> I can make a PR or include that in this PR if desired

Would be great if you could include it in this PR :)

melusc:
I have updated them all to use nullish coallescing assignments. I haven't added any more tests. I think if something changes that would undo this the test I already added should catch anything.

PR #764, opened 2024-07-28
fix: restore typescript extension support

### LINKED ISSUE #763: xo 0.59 with typescript projects fails with n/file-extension-in-import
Using `xo` >= 0.59.0 with TypeScript projects is quite useless because of [`n/file-extension-in-import`](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/file-extension-in-import.md):

```plaintext
source/index.ts:1:15
  ✖   1:15  forbid file extension .js.  n/file-extension-in-import
  ✖   2:15  forbid file extension .js.  n/file-extension-in-import
  ✖   4:15  forbid file extension .js.  n/file-extension-in-import
```

It's configured in xo here:

https://github.com/xojs/xo/blob/c96584342b41412883abeab2a63155fbddc30589/config/plugins.cjs#L314-L325

Not sure what's needed to fix this in order to update xo to the current version in TypeScript projects. `xo` 0.58 works perfectly fine.

EdJoPaTo:
fixes #763 

As far as I understand the updated rule checks for the listed extensions and only allows the listed ones now. `js` wasn't included and therefore assumed as forbidden.

As the rule was improved the reason for these custom options in gone now. It automatically assumes `js` for `ts` now.

https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/file-extension-in-import.md mentions `typescriptExtensionMap` but the default one seems to be fine so no need to configure it.

### DIFF
--- a/config/plugins.cjs
+++ b/config/plugins.cjs
@@ -315,13 +315,6 @@ module.exports = {
 		'n/file-extension-in-import': [
 			'error',
 			'always',
-			{
-				// TypeScript doesn't yet support using extensions and fails with error TS2691.
-				'.ts': 'never',
-				'.tsx': 'never',
-				'.mts': 'never',
-				'.cts': 'never',
-			},
 		],
 		'n/no-mixed-requires': [
 			'error',


fregante:
Confirmed to work. Thanks!

PR #774, opened 2025-01-14
ESLint v9 and flat config

spence-s:
My version of `xo` with flat config support. 

It's a complete re-write obviously here in TS and this is the initial PR WIP, let me know what you think. 

It seems to work pretty darn well for me so far as a drop replacement in the repos I have tested it on and it works out of the box with the vscode extension.

It also exports a `createEslintConfig` function, for users who'd rather use eslint directly but get (almost) all of the goodies from using `xo` cli directly.

The docs probably need work still and test coverage could definitely be better. 

You can test out a beta version https://www.npmjs.com/package/@spence-s/flat-xo which I will deprecate if this PR is accepted.

easy to replace in package.json `"xo": "npm:@spence-s/flat-xo@latest"`


spence-s:
I have experimented with adding a `react` flag to the beta version, Not included yet in this PR, which will enable react linting extremely easy. I really like it. 

If we feel this is something we are ok with, I will add it to this PR as well.

---

We can also discuss https://github.com/xojs/xo/issues/777 here I am open to dropping formatting support prettier while supporting prettier compatibility as an option.

---

@sindresorhus I know this is a big PR and difficult to figure out the right approach or if we want to support the new implementation or design decisions that I made. If you feel you'd like a different approach to getting this merged in or accomplished, I am open to it.

 perhaps adding the flat-config behind a flag in the current CLI would work and make it a safer change. The reason I didn't do that as step 1 is because it will be a bit odd with the JS/TS boundary.

I will try to work on getting this PR to a more "integrated" and safe state for merging, but if you have suggestions for this let me know.

fregante:
> If we feel this is something we are ok with, I will add it to this PR as well.

Given the size of the PR, I'd keep that for a follow-up instead.

lib/create-eslint-config/config.ts
```diff
@@ -0,0 +1,398 @@
...
+    /**
+     * These are the base rules that are always applied to all js and ts file types
+     */
+    rules: {
+      ...pluginAva?.configs?.['recommended']?.rules,
+      ...pluginUnicorn.configs?.recommended?.rules,
+      'no-use-extend-native/no-use-extend-native': 'error',
+      // TODO: Remove this override at some point.
+      // It's just here to ease users into readable variable names.
+      'unicorn/prevent-abbreviations': [
```
fregante, line 86: Wait until someone else confirms, but I think these are complete duplicates and can now be dropped since they appear in unicorn itself. 

The other plugins should also follow once their config is fully moved to eslint-config-xo, but that can be a follow-up (unless they're already there too)

sindresorhus:
Thanks for working on this and sorry about the slow response.

I think this is the right approach. Better to rip off the bandaid than complicate things with flags.

---

Is it not possible to support config in package.json anymore?

---

Before we eventually merge this, it would be great if you could run a script to convert indentation to tab-indentation. I could do it too, but then I would create noise for all the changes you did.

lib/create-eslint-config/config.ts
sindresorhus: Open an issue instead. It's not really related to this PR. This PR is already huge.

sindresorhus:
> I have experimented with adding a react flag to the beta version, Not included yet in this PR, which will enable react linting extremely easy. I really like it.

I'm open to it, but should definitely be in a follow-up PR.

spence-s:
>Is it not possible to support config in package.json anymore?

>Pretty much all I used it for is to disable some rules, so maybe we could support a limited set? Like only disabling rules?


Sure! I had been debating this anyway I think its a good value add for the xo cli/editor extension to be able to use package.json.

We can support any parts of the config that are serializable, so basically everything except plugins I think and some language settings maybe.

So supported config types will be:
`xo.config.js`
`xo.config.cjs`
`xo.config.cjs`
`xo.config.ts`
`xo.config.cts`
`xo.config.mts`

`package.json`

optionally it would be trivial to add `xo.config.json` as well if we wanted.

I think its nice to give users a better reason to stick with xo eco system entirely since this PR gives them an easier way to integrate xo with eslint directly.

spence-s:
> Before we eventually merge this, it would be great if you could run a script to convert indentation to tab-indentation. I could do it too, but then I would create noise for all the changes you did.

Yes, I'll take care of this as well.

sindresorhus (changes requested):

.commitlintrc.json
```diff
@@ -0,0 +1 @@
+{ "extends": ["@commitlint/config-conventional"] }
```
fregante, line 1: I see a lot of unnecessary/unrelated changes in this PR that should be reverted to speed up reviewing. The first 9 files are all unrelated. The only useful change maybe is this:

```diff
+            - 22
+            - 23
```

fregante:
As for the config, my advice is to ship a baseline XO that doesn't support much config at all, and then iterate on that.

As part of this PR, my advice would be to support the `xo` field in package.json, but just for `rules`, `overrides`, and XO flags, but not `extends`, `plugins`, and all that. You do not want to recreate eslintrc.

An `xo.config.js` file, if necessary, can come later.

spence-s:
Ok I think this is truly ready for review now 

Here is a running change log of things and important info to keep in mind for reviewing: 

## Changelog

## New Features

- Adds a static an `xoToEslintConfig` method which can be used in an `eslint.config.js` file
    - This function will not do any lookup for the user, so it will be harder to use some of the features of xo, and it will not handle auto ts lookup nor prettier lookup.
- exports the base xo config from `xo/config` ie… `import {config} from 'xo/config'` in an `eslint.config.js` file
- new CLI and config options
    - `react` - Adds a base `react` option which adds `eslint-config-xo-react` to the selected files 
    - `prettier 'compat'` Adds a new option for prettier, where if you set the option to the string `"compat"` it will simply turn off all the prettier affected @stylistic rules using `eslint-config-prettier` otherwise it works the same as it does currently.
    - Allows turning off the automatic ts handling as an optimization (it is on by default) with a new config options `ts`       
- Allows ts files as stdin `echo "const x: boolean = true" | xo --stdin --stdin-filename=stdin.ts`
- stdin option now defaults file names to stdin.js  and no longer requires the `--stdin-filename` option
- Only accepts configuration through an `xo.config.{js,cjs,mjs,ts,cts,mts}` file 
    - open to supporting other methods, but I personally only use these configs. package.json could be handled in a number of different ways if we wanted to say only support the `cli` arguments + rules object - or make it a flat config array that only supports serializable features of the config array. Happy to implement if desired. 

## Removal of Features
- Drops the node `engines` check from xo as they were only active in engines that are unsupported.
- Drops auto webpack lookup and webpack import resolution (feels a bit niche, but would be willing to support still)
- support for custom cli reporter removed (easy to add back if we wanted, didn't feel important)
- Removed CLI options
    - `plugin` no longer makes sense, still configurable with config
    - `extend` no longer makes sense, still configurable with config
    - `env` this no longer makes sense, still configurable with config
    - `global` this could potentially be added back but I have never used it from cli, still configurable with config
    - `node-version` removed because all of the node stuff was for unsupported node versions
    - `extension`  could potentially be added back, has limited value from cli, can still be configured with config
    - `reporter` this could be added back but feel like its relatively unused and requires a potentially flaky lookup open to adding this back now and will likely in the future regardless

### CLI and Change Summary
In general, this version of `xo` is much "thinner" in that we really just barely abstract over ESLint. In doing so, we make the config much more flexible for users, and really there is no feature loss when moving from `eslint` to  `xo`. The tradeoff is that the `xo` cli itself does not add as much capabilities but it currently does have a few advantages that it. 
- easily accepts ts configuration (`xo.config.ts`) with no extra steps
- will handle prettier lookups if you use that option and integration with prettier will be much smoother 
- handles stdin and stdin ts files
- automatic tsconfig resolution and typeaware linting automatically

If I think of more changes I will continue to add here, but this covers the major bases I feel. 

Also note, I have been using this for the past few months and its been working great, I think it is ready for a release, but happy to make adjustments if you see issues. 

If you'd like to test right now, my fork is published here and I'm keeping this PR and the fork synced: https://www.npmjs.com/package/@spence-s/flat-xo

To test with the fork `npm install xo@npm:@spence-s/flat-xo` and let me know if you experience issues, we already have a few beta users that have been using and the xo vscode extension is using it for linting as well as all the repos I have been working in.

sindresorhus:
> exports the base xo config from xo/config ie… import {config} from 'xo/config' in an eslint.config.js file

I don't think we should do this yet. It will require more discussion and thinking and not worth blocking this change over. We will also move more rules over to `eslint-config-xo`. You can leave everything in place, just remove the export in package.json for now.

> Allows turning off the automatic ts handling as an optimization (it is on by default) with a new config options ts

I don't think anyone has ever requested this. Is it really worth exposing an option for this?

> open to supporting other methods, but I personally only use these configs. package.json could be handled in a number of different ways if we wanted to say only support the cli arguments + rules object - or make it a flat config array that only supports serializable features of the config array.

package.json `xo` property must be supported, but we could limit it to just CLI flags, `rules`, and `ignores`.

lib/types.ts
```diff
@@ -0,0 +1,90 @@
+import type {Simplify} from 'type-fest';
+import {type ESLint, type Rule, type Linter} from 'eslint';
+
+export type Space = boolean | number | string | undefined;
+
+export type XoConfigOptions = {
+	/**
+   * Use spaces for indentation.
```
sindresorhus, line 8: The indentation is all over the place in this file.

package.json
```diff
@@ -54,68 +64,79 @@
...
-			"coverage"
-		]
+		"dedent": "^1.5.3",
+		"execa": "^9.5.2",
+		"husky": "^9.1.7",
+		"lint-staged": "^15.4.3",
+		"np": "^10.2.0",
+		"npm-package-json-lint": "^8.0.0",
+		"npm-package-json-lint-config-default": "^7.0.1",
+		"path-exists": "^5.0.0",
```
sindresorhus, line 114: This is used in the CLI, so must be in `dependencies`.

readme.md
```diff
@@ -1,17 +1,17 @@
...
-	<br>
-	<img width="400" src="media/logo.svg" alt="XO">
-	<br>
-	<br>
-	<br>
+  <br>
+  <img width="400" src="media/logo.svg" alt="XO">
+  <br>
+  <br>
+  <br>
```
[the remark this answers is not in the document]
spence-s: ahh ok - I let prettier format this, I will undo that. - also noticed some other typos in the readme to fix

sindresorhus:
Thank you so much for working on this, @spence-s. I will merge now and do a prerelease.

fregante (commented):
Finally I can post this "after merge" notes 😁

sindresorhus:
Boom: https://github.com/xojs/xo/releases/tag/v0.61.0-0

Feel free to edit release notes if I missed anything.

lib/constants.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const allExtensions = [...jsExtensions, ...tsExtensions];
+
+export const allFilesGlob = `**/*.{${allExtensions.join(',')}}`;
+
+export const moduleName = 'xo';
+
+export const tsconfigDefaults: TsConfigJsonResolved = {
+	compilerOptions: {
+		target: 'es2018',
```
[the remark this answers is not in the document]
sindresorhus: https://github.com/xojs/xo/commit/2cb0421fa646c456fee30659dc7489843edc8612#diff-561ab956a0b904914573a9544f92493c5493b6e14d30372365bcdd9c5c7f6e59L34-L42

1000ch:
@sindresorhus 
Could you republish the latest version? [0.61.0-0 is missing `dist/index.d.ts`](https://www.jsdelivr.com/package/npm/xo?tab=files&version=0.61.0-0&path=dist) built with `npm run build`. Probably `dist/*.d.ts` needs to be specified at `package.json#files`.

PR #780, opened 2025-04-03
chore: publish dts files

spence-s:
Publish d.ts files for use in editor extensions or `xo.config.{ts|js}` files.

Also a little tsconfig cleanup

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
 	},
 	"files": [
 		"dist/lib",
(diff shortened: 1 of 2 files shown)


LitoMore:
It looks we have those `dist/lib/*.js.map` files published to npm. Is that expected? If yes, do we need to add `dist/*.js.map` to `files` as well?

PR #781, opened 2025-04-03
Add await

patrik-csak:
Because `xo.xoToEslintConfig()` returns a `Promise<Linter.Config[]>`


spence-s:
we should close this and merge #790 instead, which fixes this by no longer returning a promise from `xoToEslintConfig`

PR #782, opened 2025-04-04
Fix missing declarations

LitoMore:
See https://www.jsdelivr.com/package/npm/xo?tab=files&version=0.61.0-0&path=dist

The `index.d.ts` is missing. It looks we misconfigured the `files` field in `package.json`.

I've added `dist/*.d.ts` to `files` to fix this issue.

In the meantime, we have those `dist/lib/*.js.map` files published to npm. Is that expected? If yes, do we need to add `dist/*.js.map` to `files` as well?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
 	},
 	"files": [
 		"dist/lib",
-		"dist/*.js"
+		"dist/*.js",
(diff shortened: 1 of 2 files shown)


LitoMore:
Duplicate of #780. Closing.

PR #783, opened 2025-04-04
Ignore source map files while publishing

LitoMore:
Please correct me if I'm wrong.

It looks like we only used those source map files for testing purposes. Maybe we could ignore them while publishing to npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
 		"test:setup": "node scripts/setup-tests"
 	},
 	"files": [
-		"dist/lib",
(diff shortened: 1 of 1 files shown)


sindresorhus:
I agree, I don't think we should publish source maps files. The most common usage is the CLI, and source maps add a lot of weight to the package size.

spence-s (approved):
Yep, agreed this is good. I like that this leaves them in the build as well and filters them at publish time, so test errors can be reported using source maps still.

PR #824, opened 2025-07-26
Update eslint-plugin-unicorn

justinhelmer:
See sindresorhus/eslint-plugin-unicorn#2707


justinhelmer:
This seems to be the forcing function to update some code for a new linter default, which then requires a new lib, which requires updating the dependent `@sindresorhus/tsconfig` package.

Sorry I don't have the bandwidth right now :-/ 

I will keep this open in case someone wants to run with it, or feel free to close it out if its creating noise.

PR #838, opened 2025-09-20
fix: Resolve NoFilesFoundError with glob patterns and ignores

simwai:
### Problem
XO was throwing `NoFilesFoundError: No files matching '!**/*' were found` when using glob patterns like `"*.js"` on projects where files were filtered out by ignore patterns. This was caused by double filtering - once by globby and again by ESLint.

### Root Cause
The `lintFiles` method was applying ignore patterns twice:
1. In globby (pre-filtering files)  
2. In ESLint (filtering again)

When all files were filtered out, the fallback `'!**/*'` pattern was invalid.

### Solution
- **Separate concerns**: Let globby handle system ignores (`node_modules`, etc.) and ESLint handle user ignores
- **Fix empty file handling**: Return empty result instead of invalid `'!**/*'` pattern
- **Prevent node_modules contamination**: Stops TypeScript config from including dependency files

### Testing
All 124 tests now pass, including the previously failing glob pattern tests.

**Note**: This fixes the underlying infrastructure bug but does not implement issue #238 (error messages for ignored files). That would be a separate enhancement.


lib/xo.ts
```diff
@@ -251,6 +251,30 @@ export class Xo {
 		}
 	}
 
+	/**
+	 * Checks if any directly specified files are ignored and throws an error.
+	 * @private
+	 */
+	async checkForIgnoredDirectFiles(globs: string[]): Promise<void> {
+		const {flatOptions} = await resolveXoConfig(this.linterOptions);
```
spence-s, line 259: Hmm this will cause the resolveXoConfig function to get ran 2x I think.

Instead of calling this function maybe we can acheive the same thing by doing 

```ts
await this.setCoConfig();
this.xoConfig // instead of flatOptions
```

lib/xo.ts
```diff
@@ -251,6 +251,30 @@ export class Xo {
...
+		const combinedIgnores = [
+			...defaultIgnores,
+			...arrify(this.baseXoConfig.ignores),
+			...flatOptions.flatMap(config => config.ignores ? arrify(config.ignores) : []),
```
spence-s, line 263: I don't think this is quite correct logic. If a flatConfig can has ignores AND files, then we only want to apply those ignores to the files in that config...

lib/xo.ts
```diff
@@ -251,6 +251,30 @@ export class Xo {
...
+		];
+
+		// Filter non-glob patterns first (synchronous operation)
+		const directFileGlobs = globs.filter(glob =>
+			!glob.includes('*') && !glob.includes('?') && !glob.includes('['));
```
spence-s, line 268: Couldn't direct files also have these strings? `[slug]`.js is common with next.js for instance.

test/xo/check-for-ignore-direct-files.test.ts
```diff
@@ -0,0 +1,110 @@
...
+		]
+	`;
+	await fs.writeFile(xoConfigPath, xoConfig, 'utf8');
+
+	// Run XO on first ignored file - should error on first file
+	const {stderr: s1}: ExecaError = await t.throwsAsync($`node ./dist/cli --cwd ${t.context.cwd} ignored1.js ignored2.js`);
+	t.true((s1?.toString() ?? '').includes('ignored1.js') && (s1?.toString() ?? '').includes('ignored by XO configuration'));
+});
+
+test('xo handles mixed direct files and glob patterns correctly', async t => {
```
spence-s, line 80: Where is the glob pattern in this test?

lib/xo.ts
```diff
@@ -346,8 +373,9 @@ export class Xo {
 			throw new Error('Failed to initialize ESLint');
 		}
```
spence-s, line 375: Maybe we can simplify this PR. 

Add the `combinedIgnores` to the original globby lookup to get `files`. The `combinedIgnores` needs to check whether the ignore is the only key: see https://github.com/xojs/xo/blob/main/lib/xo-to-eslint.ts#L34

Move the `checkForIgnoredDirectFiles` after the files look up and make sure it all is does is use fs.stat to filter out the direct files, then check if each direct file is included in `files` and if not, then we can throw our error.

spence-s (changes requested):

PR #845, opened 2025-10-06
Add support for Vue, Svelte, and Astro file extensions

### LINKED ISSUE #843: Additional file types cannot be enabled via config file
I was wondering why my Svelte files weren’t getting linted, even though I have a config block with `files: '**/*.svelte'`. What I discovered after much trial and error is that if I run `XO` via `xo '**/*.{js,cjs,mjs,jsx,ts,cts,mts,tsx,svelte}'` then Svelte files will get linted per the configuration in my config file; but running via just `xo` ignores them. So the command-line argument is mandatory in my case; I don’t see a way to get `XO` to lint `.svelte` files via anything I can define in `xo.config.js`.

It seems like the CLI discovers the files to lint _first,_ based on the file types defined in:

https://github.com/xojs/xo/blob/c4a13e11fcebe6ca133420cccae0e61f444cbc14/lib/constants.ts#L21-L31

and then these files are linted according to the merged-together config block(s) that apply to them. So since `.svelte` isn’t listed in `lib/constants.ts`, the config block for that file type in `xo.config.js` does nothing.

The readme discusses this somewhat:

https://github.com/xojs/xo/blob/c4a13e11fcebe6ca133420cccae0e61f444cbc14/readme.md?plain=1#L137-L142

If the current behavior is what’s desired, a more accurate explanation would be something like: “A glob or array of glob strings to which the config object will apply, as a subset of the glob that `XO` is linting per command line arguments. By default `XO` will apply the configuration to [its default glob](lib/constants.ts). If you aim to lint a file type beyond these defaults, it must be passed to `XO` on the command line such as `xo '**/*.{js,cjs,mjs,jsx,ts,cts,mts,tsx,foo}'` and a config block with `files: ['**/*.foo']` needs to be added.”

But I think more intuitive behavior would be that if `XO` discovers a config file `files` property with a glob that expands upon `XO`’s default, then the list of files it’s finding to lint gets expanded. So in other words, running simply `xo` with an `xo.config.js` config block that contains `files: '**/*.svelte'` should be enough to add `**/*.svelte` to the set of files that `XO` lints.

sindresorhus:
XO now automatically discovers and lints `.vue`, `.svelte`, and `.astro` files when running without arguments. This resolves the limitation where users had to explicitly pass these file patterns on the command line even when they were configured in the config file.

Fixes #843

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -22,11 +22,13 @@ export const tsExtensions = ['ts', 'tsx', 'cts', 'mts'];
 
 export const jsExtensions = ['js', 'jsx', 'mjs', 'cjs'];
 
+export const frameworkExtensions = ['vue', 'svelte', 'astro'];
+
 export const jsFilesGlob = `**/*.{${jsExtensions.join(',')}}`;
 
 export const tsFilesGlob = `**/*.{${tsExtensions.join(',')}}`;
 
-export const allExtensions = [...jsExtensions, ...tsExtensions];
+export const allExtensions = [...jsExtensions, ...tsExtensions, ...frameworkExtensions];
 
 export const allFilesGlob = `**/*.{${allExtensions.join(',')}}`;
 
--- a/readme.md
+++ b/readme.md
@@ -108,7 +108,7 @@ Simply run `$ npm init xo` (with any options) to add XO to create an `xo.config.
 
 ## Config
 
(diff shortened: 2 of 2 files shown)


spence-s (approved):
Yeah I am good with this! I don't use any of these frameworks but this seems reasonable and simple.

PR #858, opened 2026-01-29
Avoid duplicate plugin registration

### LINKED ISSUE #852: Cannot redefine plugin "react-hooks"
I'm trying to use XO with Next.js v16 (related: #635)

I'm getting this error when I try to add XO to a Next.js v16 project via `xoToEslintConfig()`:

> ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "react-hooks".

---

Steps to reproduce:

1. Create a Next.js project

   ```sh
   npx create-next-app@latest
   ```

1. Add XO to the project

   ```sh
   npm install xo --save-dev
   ```

1. Add XO to `eslint.config.mjs`

   ```diff
     // ...

     const eslintConfig = defineConfig([
       ...nextVitals,
       ...nextTs,
   +   ...xo.xoToEslintConfig([{ react: true }]),
       // ...
     ]);

     // ...
   ```

1. Run ESLint

   ```sh
   npm run lint
   ```

   ```
   > next-xo@0.1.0 lint
   > eslint


   Oops! Something went wrong! :(

   ESLint: 9.39.1

   ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "react-hooks".
   		at rethrowConfigError (/Users/user/next-xo/node_modules/@eslint/config-array/dist/cjs/index.cjs:343:8)
   		at /Users/user/next-xo/node_modules/@eslint/config-array/dist/cjs/index.cjs:1405:5
   		at Array.reduce (<anonymous>)
   		at FlatConfigArray.getConfigWithStatus (/Users/user/next-xo/node_modules/@eslint/config-array/dist/cjs/index.cjs:1398:43)
   		at FlatConfigArray.getConfig (/Users/user/next-xo/node_modules/@eslint/config-array/dist/cjs/index.cjs:1427:15)
   		at entryFilter (/Users/user/next-xo/node_modules/eslint/lib/eslint/eslint-helpers.js:323:27)
   		at async NodeHfs.<anonymous> (file:///Users/user/next-xo/node_modules/@humanfs/core/src/hfs.js:574:24)
   		at async NodeHfs.<anonymous> (file:///Users/user/next-xo/node_modules/@humanfs/core/src/hfs.js:604:6)
   		at async NodeHfs.walk (file:///Users/user/next-xo/node_modules/@humanfs/core/src/hfs.js:614:3)
   		at async globSearch (/Users/user/next-xo/node_modules/eslint/lib/eslint/eslint-helpers.js:364:20)
   ```

---

I'm currently getting around this by deleting XO's `react-hooks` plugin:

```js
xo.xoToEslintConfig([{ react: true }]).map((config) => {
  if (config.plugins === undefined) return config;

  const plugins = { ...config.plugins };
  delete plugins["react-hooks"];
  return { ...config, plugins };
});
```

### LINKED ISSUE #857: React compiler linting support
In preparation for adopting the react compiler, the docs [recommend](https://react.dev/learn/react-compiler/installation#eslint-integration) adding a set of lint checks. This is done by utilizing the `recommended-latest` [preset config](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md#flat-config-eslintconfigjsts):
```ts
export default defineConfig([
  reactHooks.configs.flat['recommended-latest'],
]);
```

However, when using xo's config directly:
```ts
import { type FlatXoConfig } from 'xo'
import reactPlugin from 'eslint-plugin-react'
import reactHooks from 'eslint-plugin-react-hooks';

const xoConfig: FlatXoConfig = [
    {
        react: true,
    },
    reactHooks.configs.flat['recommended-latest'], // Enable React Compiler linting rules
    reactPlugin.configs.flat['jsx-runtime'], // Enable JSX without explicit React import
]

export default xoConfig
```
I get an error:
```
$ xo .
/Users/acalvino/CodeGenerac/ic-viewit/ViewIT/clientapp/node_modules/@eslint/config-array/dist/cjs/index.cjs:343
        throw new ConfigError(configName, index, { cause: error });
              ^

ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "react-hooks".
```

Would love to be able to utilize the compiler linting rules from react-hooks, as I'm sure many others would as well; lmk know if support for that could be added, or if there's a good workaround.

### Attempted workarounds:
- disable the react option in the config - This unfortunately disables other react-related lint rules that that don't fall under react-hooks
- Just include the new react-hook rules directly under `rules` key - This results in a different error: `TypeError: Key "rules": Key "react-hooks/config": Could not find "config" in plugin "react-hooks".`

sindresorhus:
It's kinda annoying error. You don't always control the config you import.

---

Fixes #852
Fixes #857

### DIFF
--- a/lib/xo-to-eslint.ts
+++ b/lib/xo-to-eslint.ts
@@ -14,11 +14,69 @@ export type CreateConfigOptions = {
 	prettierOptions?: Options;
 };
 
+type Plugins = NonNullable<Linter.Config['plugins']>;
+type Plugin = Plugins[string];
+
+/**
+Merge all plugins from every config into a single config entry at the start of the array, ensuring user-provided plugins take precedence. This avoids ESLint's flat config rejecting duplicate plugin names.
+*/
+const hoistPlugins = (configs: Linter.Config[], userPluginOverrides: Map<string, Plugin>): Linter.Config[] => {
+	const plugins: Linter.Config['plugins'] = {};
+	const configsWithoutPlugins: Linter.Config[] = [];
+
+	for (const configItem of configs) {
+		const {plugins: configPlugins} = configItem;
+
+		if (!configPlugins) {
+			configsWithoutPlugins.push(configItem);
+			continue;
+		}
+
+		// ESLint flat config rejects duplicate plugin names, so merge all plugins into one config.
+		Object.assign(plugins, configPlugins);
+
+		const {plugins: _ignored, ...configWithoutPlugins} = configItem;
+
+		if (Object.keys(configWithoutPlugins).length > 0) {
+			configsWithoutPlugins.push(configWithoutPlugins);
+		}
+	}
+
+	for (const [pluginName, plugin] of userPluginOverrides) {
+		plugins[pluginName] = plugin;
+	}
+
+	if (Object.keys(plugins).length === 0) {
+		return configsWithoutPlugins;
+	}
+
+	return [
+		{
+			name: 'xo/plugins',
+			plugins,
+		},
+		...configsWithoutPlugins,
+	];
+};
+
 /**
 Takes a XO flat config and returns an ESlint flat config.
 */
 export function xoToEslintConfig(flatXoConfig: XoConfigItem[] | undefined, {prettierOptions = {}}: CreateConfigOptions = {}): Linter.Config[] {
 	const baseConfig = [...config];
+	const userPluginOverrides = new Map<string, Plugin>();
+
+	for (const xoConfigItem of flatXoConfig ?? []) {
+		const {plugins} = xoConfigItem;
+
+		if (!plugins) {
+			continue;
+		}
+
+		for (const [pluginName, plugin] of Object.entries(plugins)) {
+			userPluginOverrides.set(pluginName, plugin);
+		}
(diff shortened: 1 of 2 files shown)


spence-s (approved):
I think this is a great approach to hoist these. 

But 1 callout is that, in our config.ts file, we define the @typescript-eslint plugin for both javascript and typescript configs separately. ([ln 47](https://github.com/xojs/xo/blob/main/lib/config.ts#L47) and [ln 384](https://github.com/xojs/xo/blob/main/lib/config.ts#L384) respectively)

It's been a while, but I am wanting to say I did that for a reason... I am not sure if it matters or not anymore, but I think this will hoist that to the top as well. It honestly may not be a problem at all, but just a call out. 

Since all the tests are passing I am not too concerned, but it is something to watch out for.

Another call out is potential for plugins that lint specific non js/ts file types, like a JSON or MD or CSS plugin, would those behave weirdly when hoisted to the top? I am not sure, I have never tried to add those specifically and we filter those out from `eslint-config-xo` might should add some tests for these cases.

I really like the approach though and think we should try to move forward once we're sure those edge cases will work fine.

PR #876, opened 2026-04-02
Add `xo/eslint-adapter` and remove `xoToEslintConfig`

### LINKED ISSUE #875: Revisit or remove `xoToEslintConfig`
What's the use case for this utility now that plugins have been moved to the ESLint-compatible config?

- direct XO cli users don't need it
- ESLint users should use `eslint-config-xo`
- XO cli users that want IDE integration should not be able to further customize XO inline *just for the IDE*

## Suggested changes

Since only the last use case makes sense (AFAIK), customization should be removed from this helper:

- Remove function parameter
- Change usage to `export {default} from 'xo/eslint-adapter'`

The adapter should then just read XO's config from its native config locations (xo.config.ts, package.json)

sindresorhus:
Fixes #875

### DIFF
--- a/lib/xo-to-eslint.ts
+++ b/lib/xo-to-eslint.ts
@@ -83,6 +83,7 @@ const hoistPlugins = (configs: Linter.Config[], userPluginOverrides: Map<string,
 	];
 };
 
+// TODO: Remove this export if no one complains. Users should use `xo/eslint-adapter` instead.
 /**
 Takes a XO flat config and returns an ESlint flat config.
 */
--- a/lib/xo-to-eslint.ts
+++ b/lib/xo-to-eslint.ts
@@ -83,11 +83,8 @@ const hoistPlugins = (configs: Linter.Config[], userPluginOverrides: Map<string,
 	];
 };
 
-// TODO: Remove this export if no one complains. Users should use `xo/eslint-adapter` instead.
 /**
-Takes a XO flat config and returns an ESlint flat config.
-
-@deprecated
+Takes a XO flat config and returns an ESLint flat config.
 */
 export function xoToEslintConfig(flatXoConfig: XoConfigItem[] | undefined, {prettierOptions = {}}: CreateConfigOptions = {}): Linter.Config[] {
 	const baseConfig = [...config];
--- a/lib/xo-to-eslint.ts
+++ b/lib/xo-to-eslint.ts
@@ -86,6 +86,8 @@ const hoistPlugins = (configs: Linter.Config[], userPluginOverrides: Map<string,
 // TODO: Remove this export if no one complains. Users should use `xo/eslint-adapter` instead.
 /**
 Takes a XO flat config and returns an ESlint flat config.
+
+@deprecated
 */
 export function xoToEslintConfig(flatXoConfig: XoConfigItem[] | undefined, {prettierOptions = {}}: CreateConfigOptions = {}): Linter.Config[] {
 	const baseConfig = [...config];
--- a/readme.md
+++ b/readme.md
@@ -312,20 +312,16 @@ You can opt out of XO's automatic tsconfig handling by specifying your own `lang
 
 ## Usage as an ESLint Configuration
 
-With the introduction of the ESLint flat config, many of the original goals of `xo` were brought into the ESLint core, and shareable configs with plugins became possible. Although we highly recommend the use of the `xo` cli, we understand that some teams need to rely on ESLint directly.
+If you want to use ESLint directly without the `xo` CLI, use [`eslint-config-xo`](https://github.com/xojs/eslint-config-xo).
 
-For these purposes, you can still get most of the features of `xo` by using our ESLint configuration helpers.
+### ESLint adapter
 
-### xoToEslintConfig
-
-The `xoToEslintConfig` function is designed for use in an `eslint.config.js` file. It is NOT for use in an `xo.config.js` file. This function takes a `FlatXoConfig` and outputs an ESLint config object. This function will neither be able to automatically handle TS integration for you nor automatic Prettier integration. You are responsible for configuring your other tools appropriately. The `xo` cli, will, however, handle all of these details for you.
+If you already use the `xo` CLI and want your editor's ESLint integration to match exactly, use the `xo/eslint-adapter` subpath export. It automatically reads your `xo.config.js` and converts it to an ESLint flat config.
 
 `eslint.config.js`
 
 ```js
-import xo from 'xo';
-
-export default xo.xoToEslintConfig([{space: true, prettier: 'compat'}]);
+export {default} from 'xo/eslint-adapter';
 ```
 
 ## Tips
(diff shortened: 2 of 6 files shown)


lib/xo-to-eslint.ts
fregante, line 89: Should this be marked `@deprecated` too? I'm sure people won't notice this change (and complain) until it's already removed.

fregante (approved):
🙌

readme.md
```diff
@@ -312,20 +312,16 @@ You can opt out of XO's automatic tsconfig handling by specifying your own `lang
...
+If you want to use ESLint directly without the `xo` CLI, use [`eslint-config-xo`](https://github.com/xojs/eslint-config-xo).
 
-For these purposes, you can still get most of the features of `xo` by using our ESLint configuration helpers.
+### ESLint adapter
 
-### xoToEslintConfig
-
-The `xoToEslintConfig` function is designed for use in an `eslint.config.js` file. It is NOT for use in an `xo.config.js` file. This function takes a `FlatXoConfig` and outputs an ESLint config object. This function will neither be able to automatically handle TS integration for you nor automatic Prettier integration. You are responsible for configuring your other tools appropriately. The `xo` cli, will, however, handle all of these details for you.
+If you already use the `xo` CLI and want your editor's ESLint integration to match exactly, use the `xo/eslint-adapter` subpath export. It automatically reads your `xo.config.js` and converts it to an ESLint flat config.
```
spence-s: This explanation doesn't really make any sense to me. What does the adapter do again?

lib/xo-to-eslint.ts
sindresorhus: Actually, no. This method is not accessible to end-users.

readme.md
sindresorhus: I have tried to clarify it: https://github.com/xojs/xo/pull/876/commits/404fad4266c6e889988b1d98a7243b3253390e40

spence-s: Interesting...is your idea to move away from from our own editor integrations entirely?

Makes sense, although I don't love the idea of having to have an xo and an eslint config. 

Will this adapter work as a base for a shareable eslint config? LIke if I create an npm package that just epxorts this adapter, will it work like a normal shareable eslint config? --- seems like it, but want to clarify.

spence-s: To me it seems like it does the same thing as xoToEslintConfig, but with extra steps now as I have to define my xo config separately from my eslint config....

In my opinon this is a step backwards from a just a simple function (`xoToEslintConfig`) and now has file resolution logic tightly tied in...

Is that all correct, what is the value add from your point of view of this feature?

fregante: See the original request issue. 

In short, if you want to use eslint you should use eslint-config-xo directly. 

As the name suggests, this is just a compatibility layer for those who want it or need it. There are a number of services that support eslint but would never support xo.config.ts. This export is for them.

fregante: To clarify further:

- ESLint cli users use `eslint-config-xo`
- XO cli users use xo.config.ts
  - the subset of these users who *also* want broader ESLint compatibility can use this new export, which is a compatibility layer and nothing more.

spence-s:
I just wanted to bring this to top level as a comment:

I generally feel this change is a step backwards for xo. It simply makes it more obtuse to get prettier integration and react integration. I think file look ups and internal resolutions is always a brittle design (may work perfectly IDK, but that's besides the point).

Exporting a function was dead simple in comparison and was much more flexible.

The use case is also kind of silly, if you want editor integration with the cli, you should use our editor integrations. They already accomplish the goal of getting your editor to match your cli in a much more graceful way.

---
Pros: 
- More clearly separates concerns between the config and the cli (sort of there is still react/prettier/ts integration which are not included in the eslint config, all of which I use) 

Cons: 
- <s>adds extra internal steps that are non-trivial to look up more files on each run.</s> I think I am wrong about this, unsure, but it probably doesn't affect the xo cli runs?
- Similar to the pro, it still doens't really separate concerns through all the config. I think the xo package jumped in popularity (from about 90k to close to 300k weekly dls shortly after the release of v1) mainly because of xoToEslint and because of the react integration. 
- Will force some users (me and my teams) to now have 2 configuration files to maintain instead of 1 to have the same abilities...

Anyway, at best this PR is taking away some flexibility and not giving anything new.

At worst this PR is taking away some flexibility, and not giving anything new.

Why do it?

--- 

As an alternative that I would fully support: I would propose we leave this adapter out entirely and put the full functionality of xoToEslintConfig into eslint-config-xo. 

I think it will take a few more boolean options, and achieve the same flexibility as xoToEslint 

`xo` cli then becomes a very thin wrapper over eslint and eslint-config-xo that provides automatic typescript integration (in the absence of a tsconfig).

Would be simple to provide a migration guide for every use case that way that doesn't _feel_ different to users.

spence-s:
Another `eslint` regression that will come from this is that editor integrations watch and detect changes in their config files...

IE eslint editor integration looks for changes in eslint config files and re-lints when changes occur.

It will no longer work like that for users of the adapter, they will need to manually restart eslint in the editor everytime a change is made to the xo config.

`xoToEslintConfig` supported this naturally, as the config just lived in the `eslint.config` file. Its just another gotchya to the adapter method. 

Not sure if this same thing would affect the eslint cli config caching. Not really sure if that occurs, but if it does somehow, or if they add that in the future, it would have similar issues.

---

I would totally support simple renaming `xoToEslintConfig` to `xo.eslintAdapter()` as a function as well :)

fregante:
> `xo` cli then becomes a very thin wrapper over eslint and eslint-config-xo that provides automatic typescript integration (in the absence of a tsconfig).

Hello! I've been advocating for this [for years!]( https://github.com/xojs/eslint-config-xo/pull/86#:~:text=eslint%2Dconfig%2Dxo%20contains%20all%20the%20logic%20XO%20needs%2C%20and%20XO%20becomes%20a%20wrapper%20around%20eslint%20and%20eslint%2Dconfig%2Dxo) 😁 

> IE eslint editor integration looks for changes in eslint config files and re-lints when changes occur.

It seems we're stuck on the misunderstanding of who this layer is for. Let me try again: if you do not use XO's CLI and IDE integration, you do not need this layer; use `eslint-config-xo`

Now, if `eslint-config-xo` doesn't do everything xo can, that's another story. It's what https://github.com/xojs/xo/issues/769 was about.

spence-s:
Yes - I think its a good point that the function `xoToEslintConfig` does provide a way for folks to misuse xo cli. Although we clearly state they shouldn't be used together. I haven't given this much thought or seen any issues that this was happening.

Here is my gripe example: 

I am using eslint and and eslint editor integration only. 

My `eslint.config.js` is:
```js
import xo from 'xo'

export default xo.xoToEslintConfig([{react: true, prettier: true}, ....otherOverrides ]);
``` 

With this change, I now have to add an `xo.config.js`

```js
export default [{react: true, prettier: true}, ...anyOtherOverrides]
```
AND I have to keep my eslint.config.js

```js
import config from 'xo/eslint-adapter'

export default config;
```
---

This is confusing and a step backwards in my opinon. 

BUT I would be totally happy if `eslint-config-xo` supported the react and prettier options, which it does not. 

I will be stuck in a weird in between where both my `xo` editor extension will be activated by the presence of an xo config AND the eslint extension will be activated by the eslint config and in addition, eslint extension will no longer respond to my changes like it normally would as those will now change in the xo config....

 We are leaving the react and prettier users (me almost always, and likely a significant portion of other xo users who use xoToEslintConfig and prettier or react integration) in a bad spot with this change I feel.

Its a worst of all world in my opinion.

fregante:
> I would be totally happy if eslint-config-xo supported the react and prettier options

Yes, in my view `eslint-config-xo` doesn't have any reason to lack configuration/plugins present in the main `xo` package. If prettier and react are light enough to be part of `xo`, then they're surely welcome in `eslint-config-xo` too. 

This means more work is needed before `xoToEslintConfig` is removed.

PR #883, opened 2026-05-17
Fix parent XO config resolution from nested package cwd

### LINKED ISSUE #733: Projects in a Turborepo monorepo not using workspace package.json config
Hello,

I have a very simple XO configuration I'd like to share between all the packages and apps in my Turborepo monorepo:

```json
"xo": {
  "prettier": true,
  "rules": {
    "import/extensions": 0
  }
}
```

The project README states that I should just need one version of this in the root package.json file:

- https://github.com/xojs/xo#monorepo

However, I'm seeing linting errors in each of my packages if I don't have this XO config in every package.json. Have I configured something incorrectly?

sjh9714:
## Summary

Fixes XO config resolution when running from a nested package in a monorepo.

## What changed

- Added regression coverage for resolving a parent `package.json` XO config from a nested package `cwd`.
- Updated config discovery to search parent project directories instead of stopping at the package parent directory.

## Why

Issue #733 reports that packages in a monorepo do not pick up the root XO config. The previous search stopped before reaching the monorepo root.

## Testing

- `npm run build`
- `npm run test:setup`
- `npx ava dist/test/resolve-config.test.js`
- `node dist/cli.js lib/resolve-config.ts test/resolve-config.test.ts`
- `git diff --check`

Note: I also checked `npm test`; it currently stops at the lint step on both this branch and `upstream/main` with the same pre-existing `@typescript-eslint/no-unnecessary-type-assertion` reports in `lib/utils.ts` and `lib/xo.ts`.

Fixes #733

### DIFF
--- a/lib/resolve-config.ts
+++ b/lib/resolve-config.ts
@@ -24,9 +24,8 @@ export async function resolveXoConfig(options: LinterOptions): Promise<{
 			options.cwd = path.resolve(process.cwd(), options.cwd);
 		}
 
-		const stopDirectory = path.dirname(options.cwd);
-
 		const flatConfigExplorer = cosmiconfig(moduleName, {
+			searchStrategy: 'project',
 			searchPlaces: [
 				'package.json',
 				`${moduleName}.config.js`,
@@ -38,7 +37,6 @@ export async function resolveXoConfig(options: LinterOptions): Promise<{
 				'.ts': loadTypeScriptConfig,
 				'.mts': loadTypeScriptConfig,
 			},
-			stopDir: stopDirectory,
 			cache: true,
 		});
 
--- a/lib/resolve-config.ts
+++ b/lib/resolve-config.ts
@@ -1,6 +1,8 @@
+import fs from 'node:fs/promises';
 import path from 'node:path';
 import process from 'node:process';
 import {cosmiconfig} from 'cosmiconfig';
+import micromatch from 'micromatch';
 import arrify from 'arrify';
 import {createJiti} from 'jiti';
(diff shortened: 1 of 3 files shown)


sindresorhus:
I think this is still a bit risky. The fix depends on `searchStrategy: 'project'` walking past a nested `package.json` to find the parent config, but cosmiconfig documents that strategy as stopping once it finds a `package.{json,yaml}`. It happens to work with the current implementation, but if cosmiconfig changes to match its docs, this regression comes back.

It also means XO can now pick up an unrelated ancestor config above the actual project, which feels broader than the issue needs. I think we should either avoid relying on this undocumented behavior, or add an explicit test that shows this wider ancestor lookup is intentional.

I suggest opening an issue on cosmicconfig to clarify.

sjh9714:
Thanks, agreed. I removed the `searchStrategy: 'project'` change and restored the bounded cosmiconfig search.

The follow-up now uses an explicit fallback only for a parent `package.json` that declares `workspaces` matching the nested `cwd`, so XO can pick up the intended monorepo root config without relying on cosmiconfig walking past a nested `package.json` or grabbing an unrelated ancestor config.

I also added a regression test for the negative case where an ancestor `package.json` has `xo` config but does not declare the nested package as a workspace.

Ran:
- `npm run build`
- `npm run test:setup && npx ava dist/test/resolve-config.test.js`
- `node dist/cli.js lib/resolve-config.ts test/resolve-config.test.ts`
- `npm test`
- `git diff --check`

I did not open a cosmiconfig issue for now because this PR no longer depends on that undocumented behavior.

sindresorhus:
I'm not interested in maintaining this much extra code for this use-case. I recommend opening a PR to cosmicconfig instead.

PR #893, opened 2026-07-05
Update ISSUE_TEMPLATE to remove TypeScript link

fregante:
Archived: https://github.com/xojs/eslint-config-xo-typescript/issues


fregante:
Thanks Claude, it took forever to get right

